PR #77, opened 2013-05-16
fixing usage of DatagramSocket

Savar:
The DatagramSocket isn't working like the Socket class. The Socket
constructor is taking the _remote_ address and _remote_ port where
it should connect to. The DatagramSocket constructor is taking the
_local_ host/ip and _local_ port where it should bind to.

See:
http://docs.oracle.com/javase/7/docs/api/java/net/DatagramSocket.html#constructor_summary
http://docs.oracle.com/javase/7/docs/api/java/net/Socket.html#constructor_summary

I'm not able to test it easily because the StatsDWriter looks as if it
is bypassing the pool completely and nobody else seems to use it. 
At least in an older version of JMXtrans (201210..) we got exceptions
about "java.net.BindException: Cannot assign requested address".

Please take a look if this looks fine!


gehel:
Finally took time to rebase this PR. New version available in #225.

PR #207, opened 2015-01-05
map support

crysmi:
I have a jmx enabled application that returns a Map object as the attribute value and the final commit here simply passes the Map object to the getNewResultObject method

The other two commits reflect changes I needed in order to run the jar:
compiled on Windows 7 Java jdk1.7.0_09 mvn  3.0.5 
and run on CentOS release 6.6 (Final) java-1.7.0-openjdk-devel-1.7.0.71-2.5.3.2.el6_6.x86_64

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -594,6 +594,12 @@
 						<configuration>
 							<filters>
 								<filter>
+						<!-- 
+							Signature files from other jars, if included, prevent the "all" jar from starting with:
+							"java.lang.SecurityException: Invalid signature file digest for Manifest main attributes"
+							This filter excludes signatures files from other signed jars.
+							Specifically avoid including ECLISPE_.* from org.eclipse.jgit.
+						-->
 										<artifact>*:*</artifact>
 										<excludes>
 											<exclude>META-INF/*.SF</exclude>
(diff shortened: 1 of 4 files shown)


gehel:
Thanks for the PR ! I'd like to have your feedback on the comments I made before merging. Feel free to tell me I'm wrong ;-)

crysmi:
Ill add the inline comment for the shade filter but I want to check if an invalid signature was the issue first and see which files I was actually excluding

crysmi:
Upfront, apologies for my ignorance...

So the two files the filter excludes are:
META-INF/ECLIPSE_.RSA
META-INF/ECLIPSE_.SF

and they seem to be coming from the jar for org.eclipse.jgit

I added the filter after various pointers around the net led me here:
http://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html

Im not sure I understand anymore why I needed to exclude the signature - I had thought it was because the build flattened the included jars into one, therefore the signature included from jgit no longer matched.
Is no-one else having this problem?
Is the general wildcard form of the filter (ie META-INF/*.RSA, META_INF/*.SF) acceptable?

gehel:
Thanks a lot !

PR #209, opened 2015-01-12
use logback config for logger

crysmi:
Dont know when the log4j logger stopped working but Ive only been getting output in jmxtrans..out with no file rotation 
Dropped the log4j.xml and replaced with logback.xml (which should be similar logging levels)
Added the ability fo ruser to override the logback config by specifying their own config file location
the default config now zips the daily file and keeps 7 days back


gehel:
That was probably caused by a reordering of dependencies. We have dependency on both logback and log4j (there is a log4j outputwriter). I'm wondering if it is better to move to logback and update configuration files or make sure log4j works as it did at some point. Keeping the same behaviour has the obvious advantage of not requiring any change of configuration for all current deployments.

Side note : we probably need a [macker](https://innig.net/macker/) rule to ensure that all logging is coherent as we have multiple logging frameworks available at build time.

crysmi:
was just looking at moving the keyoutwriter to logback with the hope of removing log4j altogether in favour of logback - then remove the slf4j-log4j from the pom?
any reason beyond keyoutwriter to keep log4j?

gehel:
The only reason I can see is backward compatibility. I'd like @lookfirst advice on this one.

crysmi:
going with "not requiring any change of configuration for all current deployments" makes a lot of sense - pulling slf4j.impl out of keyoutwriter looks a bit more complicated too
reject this pull request and ill go for getting log4j running again

lookfirst:
No reason to keep log4j. I'd rather that everything go to the interfaces of something like slf4j so we can put whatever output we want on things.

PR #210, opened 2015-01-12
reorder pom with slf4j-log4j12 at top

crysmi:
add ignore for sortpom
simpler and more coherent fix to go back to log4j logging

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,14 @@
 
 	<dependencies>
 
+		<?SORTPOM IGNORE?>
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-log4j12</artifactId>
+			<version>1.7.7</version>
+		</dependency>
+		<?SORTPOM RESUME?>
+
 		<dependency>
 			<groupId>ch.qos.logback</groupId>
 			<artifactId>logback-classic</artifactId>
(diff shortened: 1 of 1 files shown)


gehel:
I did not test your PR at all for the moment, but can you check that this works also in the context of the shaded jar and the service wrapper ? Also, not sure of the top of my head, but is the classpath garanteed to be order dependant ? And last question (for the moment), if we do not have any direct dependency on logback (which *should* be the case), can we make sure it does not get loaded at all ?

crysmi:
Thanks gehel, hold off on any testing right now... I think I jumped a bit quick there

Maven 2 included a change to deterministic classpath ordering of dependencies (based on the order in the pom). dont know if this is still the case. SLF4j site suggest classpath may not always be obeyed in initialization ?!?!

It works in my tests of the shaded jar but the travis build is so full of slf4j "multiple binding" warnings, its now way too long to read

logback is used by TimeBasedRollingKeyOutWriter and it makes specific imports of 
logback-classic so it cant be removed without changing this code - i can have a look tomorrow

gehel:
We should *probably* remove the direct dependency from `TimeBasedRollingKeyOutWriter` to logback and have it go through SLF4J as well. Then is should be fairly easy to have a single SLF4J binding (and use log4j consistently).

That also point to the problem of having OutputWriters implementation in the same module as the core JmxTrans, they bring a whole lot of dependencies with them ...

If you have some free time, you might be interested in looking to [JmxTrans 2](https://github.com/jmxtrans/jmxtrans2) as well. I'm trying to rewrite it, merging the different JmxTrans projects (agen, embeded, standalone) in a single project.

gehel:
As @lookfirst commented on #209, it make a lot of sense to use SLF4J everywhere and choose which backend to use at assembly time.

gehel:
Sorry, I let that one slip for too long ...

While reordering dependencies is not sufficient to solve all of our problems, it will clean up the build logs.

PR #213, opened 2015-01-13
Fix for jmxtrans losing connection to Graphite and never reconnecting

jacklund:
We found issues using jmxtrans with Graphite where jmxtrans would lose its connection to Graphite and never reconnect. This PR fixes this issue (at least for us).

### DIFF
--- a/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
+++ b/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
@@ -101,6 +101,11 @@ public class GraphiteWriter extends BaseOutputWriter {
 							log.debug("Graphite Message: {}", line);
 							writer.write(line);
 							writer.flush();
+							if (writer.checkError()) {
+								log.error("Error writing to Graphite, clearing Graphite socket pool");
+								pool.returnObject(address, socket);
+								pool.clear();
+							}
 						} else {
 							log.warn("Unable to submit non-numeric value to Graphite: [{}] from result [{}]", value, result);
 						}
--- a/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
+++ b/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
@@ -101,6 +101,7 @@ public class GraphiteWriter extends BaseOutputWriter {
 									+ result.getEpoch() / 1000 + "\n";
 							log.debug("Graphite Message: {}", line);
 							writer.write(line);
+							writer.flush();
 						} else {
 							log.warn("Unable to submit non-numeric value to Graphite: [{}] from result [{}]", value, result);
 						}
@@ -108,7 +109,7 @@ public class GraphiteWriter extends BaseOutputWriter {
 				}
 			}
 		} finally {
-			if (writer.checkError()) {
+			if (writer != null && writer.checkError()) {
 				log.error("Error writing to Graphite, clearing Graphite socket pool");
 				pool.invalidateObject(address, socket);
 			} else {
--- a/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
+++ b/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
@@ -105,6 +105,8 @@ public class GraphiteWriter extends BaseOutputWriter {
 								log.error("Error writing to Graphite, clearing Graphite socket pool");
 								pool.returnObject(address, socket);
 								pool.clear();
+								socket = pool.borrowObject(address);
+								writer = new PrintWriter(new OutputStreamWriter(socket.getOutputStream(), UTF_8), true);
 							}
 						} else {
 							log.warn("Unable to submit non-numeric value to Graphite: [{}] from result [{}]", value, result);
--- a/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
+++ b/src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
@@ -80,10 +80,11 @@ public class GraphiteWriter extends BaseOutputWriter {
 
 	public void internalWrite(Server server, Query query, ImmutableList<Result> results) throws Exception {
 		Socket socket = null;
+		PrintWriter writer = null;
 
 		try {
 			socket = pool.borrowObject(address);
-			PrintWriter writer = new PrintWriter(new OutputStreamWriter(socket.getOutputStream(), UTF_8), true);
+			writer = new PrintWriter(new OutputStreamWriter(socket.getOutputStream(), UTF_8), true);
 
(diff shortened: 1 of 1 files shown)


src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
```diff
@@ -101,6 +101,11 @@ public void internalWrite(Server server, Query query, ImmutableList<Result> resu
 							log.debug("Graphite Message: {}", line);
 							writer.write(line);
 							writer.flush();
+							if (writer.checkError()) {
+								log.error("Error writing to Graphite, clearing Graphite socket pool");
+								pool.returnObject(address, socket);
+								pool.clear();
```
gehel: Clearing the pool to get rid of a single dirty connection is a bit overkill. It is also non thread safe as there is a chance that the dirty connection is out of the pool again when we clear it (and `clear()` only clears the idle objects. Using `invalidateObject()` is probably better.

That being said, good catch ! We've been having the same kind of problems, but could not pinpoint the cause !

gehel:
Would be great to have a unit test for this. Not the easiest to write, but a fairly good exercise ...

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
jacklund: True, but we weren't sure if the other connections were bad as well (for instance, if it was a problem with Graphite dropping the connections), and since this is a production application for us, we felt like taking the hit and reinitializing all the connections would be worth it. The other aspect of this is that it was intermittent, so we'd have to wait for a day or so after deploying a fix to make sure it actually worked, and we needed it fixed ASAP.

jacklund:
Believe me, if we could have, we would have. We tried reproducing it by running jmxtrans and Graphite on separate boxes, then setting iptables on the Graphite box to cause the connection to drop without restarting Graphite, but it didn't reproduce it. We think that there's probably a race condition such that in most cases it reconnects, but in a small percentage of the cases it doesn't. Kind of a hard thing to unit test. :-)

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
```diff
@@ -101,6 +101,11 @@ public void internalWrite(Server server, Query query, ImmutableList<Result> resu
 							log.debug("Graphite Message: {}", line);
 							writer.write(line);
 							writer.flush();
```
gehel: Additional note, this `flush()` should be moved at the end of the treatment to ensure optimal use of buffers

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
```diff
@@ -101,6 +101,11 @@ public void internalWrite(Server server, Query query, ImmutableList<Result> resu
...
+							if (writer.checkError()) {
+								log.error("Error writing to Graphite, clearing Graphite socket pool");
+								pool.returnObject(address, socket);
```
gehel: Object is returned to pool but reused later in the loop, not good ...

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
jacklund: Don't think that will work. The `flush()` is what causes the error to emerge. `write()` just fills the datagram, but it doesn't get sent until the `flush()`. If we put the `flush()` later, our check will always show as false.

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
jacklund: Good call. We'll fix.

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
mnussbaum: I think the idea is it's better to be overzealous and clear the entire pool, since it incurs minimal overhead and ensures that future writes will succeed. If one persistent connection is dead the other connections have suffered through the same issue and there is a good chance they are dead as well.

As for the thread safety, if another connection is checked out and the write succeeds then it's all good, if it fails then it will trigger the connection reset again and be good to go on the next write.

In a pathological case where every connection is checked out we would lose as many datagrams as there are connections, but even this seems not too bad compared to further complicating the code.

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
gehel: Actually, your implementation is robust. You do get a new connection after returning the faulty one. Still, we should use `invalidateObject()` instead of `clear()`.

src/com/googlecode/jmxtrans/model/output/GraphiteWriter.java
gehel: Nope, the `checkError()` will call `flush()` so you are good with just calling `checkError()`.

gehel: The real question is performance vs data loss. If we checkError / flush for each metric, we ensure that problems are detected early and corrected as soon as possible. If we check late, we ensure optimal use of buffers and less pressure on network. I would tend to go with performance instead of consistency. I think loosing a few data point is OK if we can ensure high throughput.

jacklund:
So, we went ahead and moved the check outside the loop, and used `invalidateObject()`. I think that should still do what we want, although in the case where all the connections are bad it will probably lose a very small amount of data per connection, but that seems like a reasonable tradeoff.

gehel:
That's looking good ! Thanks for the rework !

Still a few questions / considerations ... You completely removed the `flush()`, which is good for performance, but then there is no garantee that the socket will ever be flushed. There is no garantee that the socket will ever be reused. I think we should still flush after a batch of results.

Second point is really a detail, but technically, the writer could be uninitialized if exception is thrown when we get the output stream of the socket, so there is a slight chance to get a NPE when checking for errors and thus not invalidating the socket. Honestly, I'm not sure if there is a real use case where that could happen, so checking for it might be overkill ...

And last point, you could write a test for that by mocking the socket. That could be a variation of the `GraphiteWriterTests.writeSingleResult()` test. You've done more than enough by finding the bug and fixing it, if you dont write the test, I will ...

gehel:
You're wonderfull ! And I'm not (or too tired ...). `checkErrors()` will flush, so no need to flush before. And flushing should be done only at the end to use buffering as much as possible.

I'll merge and push a fix afterward...

gehel:
Just for info, test added and flush removed, everything pushed to master ...

jacklund:
Thanks Guillaume!

PR #218, opened 2015-01-27
Fix rpm build

beevee:
1. Do not attempt to create `/usr/bin` directory to avoid conflict with `filesystem` package on RHEL7 / CentOS7. See #176 (fixed the same problem for ant build).
2. Include `tools` directory with `yaml2jmxtrans.py` script.


gehel:
Looks good, thanks a lot !

PR #219, opened 2015-01-27
Add allowDottedKeys option to config converter

beevee:
This recent feature is missing from `yaml2jmxtrans.py` script.


gehel:
Looks good too, thanks a lot !

Issue #221, opened 2015-02-03
logging is not working as expected


praste:
I am not sure why are we redirecting process output to the log file in 'start()' function inside 'jmxtrans.sh' script.

I see output always going to stdout (even when I commented it out). If I am using StdoutWriter, I would like to separate jmxtrans logs from query results

gehel:
StdoutWriter is probably not the best OutputWriter for a production setup. It is quite useful for testing purpose. I suggest you have a look at the [Log4JWriter](https://github.com/jmxtrans/jmxtrans/blob/master/src/com/googlecode/jmxtrans/model/output/Log4JWriter.java) to route metrics to a specific file. We might want to implement a dedicated FileOutputWriter ...

While JmxTrans logs should go to a file (you can tune logging configuration for your specific needs) there are some cases where you will still get garbage on stdout (the first being error in the setup of the logging framework).

Does that make sense for you ?

praste:
I am using StdoutWriter for testing only. What I am wondering about is if
log file is already specified in log4j.xml then why is there a need to
redirect output of java process. If I remove the redirect I don't see
output being written to the file specified in log4j. Also, changing
LOG_LEVEL to info etc has no effect.


On Tue, Feb 3, 2015 at 5:26 PM, Guillaume Lederrey <notifications@github.com

gehel:
Redirecting output when starting a service is a good practice. If you dont and there is some output (which will happen in some error cases) you'll loose that output.

Note that there might well be some problems in the current log4j configuration. #215 has been opened about similar problems. Some cleanup is required ...

If you have some free time, a pull request is always welcomed !

praste:
I did some digging and looks like as we have added logback to our dependencies, it now expects logback.xml instead of log4j.xml (I am not sure if there is a way to make log4j.xml to work with current library dependencies).

Even if it is possible to do so, we should either 'add classpath attribute to the manifest and explicitly add log4j.xml to the classpath or move log4j.xml outside the jar and add -Dlog4j.configuration=<path_to_log4j.xml> to VM arguments.

I will submit a patch for logback, however, I think most of us would hate to move to logback

gehel:
As discussed in #215, the preferred solution would be to move everything we can to slf4j and make the choice of logback vs log4j as late as possible. But just having a correctly working logging system, whatever the solution, is of course a good first step anyway.

stevenbower:
[...]

PR #223, opened 2015-02-12
Use logback instead of log4j to fix logging issues

praste:
This fixes issue #221 . 
We have both `log4j` and `logback` jars and looks like `slf4j` is hitting `logback` jar  first(as classloader seems to be loading those before `log4j`). In this case `logback` ignore `log4j.xml` and uses default configuration where it spits out everything to console. Default console logger is synchronous and hence it could be performance issue as well if we are logging too much. 

`Logback` has really cool features like ability to change log level at runtime using JMX (which is available in `log4j2` but not in `log4j`)

I have put together `logback.xml` to mimic intending log configuration we were trying to achieve using `log4j.xml`, with some more added features like automatically delete log file more than 15 days old.


gehel:
Looks good. And seeing the current state of logging in JmxTrans, it can only improve the situation ;-)

gehel:
Thanks for your help !

jmxtrans.sh
```diff
@@ -17,7 +17,7 @@ fi
 
 JAVA_HOME=${JAVA_HOME:-"/usr"}
 LOG_DIR=${LOG_DIR:-"."}
-LOG_FILE=${LOG_FILE:-"/dev/null"}
+STDOUT_REDIRECT_FILE=${STDOUT_REDIRECT_FILE:-"/dev/null"}
```
lookfirst: breaks backwards compat

gehel: Damn me ... I'll revert right now ...

gehel:
Note that the move from log4j.xml to logback.xml is also a breaking compatibility, but it actually fixes the log4j / logback conflict that is breaking logging for a fairly long time ...

My opinion: we should break compatibility and make logging work ...

jmxtrans.sh
praste: You don't really need a LOG_FILE variable as log file name was already
there in lo4j.xml

Using that environment variable was hack to make broken logging working.

gehel: Naming might not be the best ever, but it's not a hack, it's a good practice to capture stdout on a daemon process. You can think about it as 2 different levels of logging.

lookfirst: Agreed with @gehel

Issue #215, opened 2015-01-16
Logging implementation may be broken


crysmi:
Due to the dependency on logback-classic (TimeBasedRollingKeyOutWriter) and slf4j-log4j (KeyOutWriter), both of these libraries are included at runtime. This is not recommended (see http://www.slf4j.org/codes.html#multiple_bindings)

This has had multiple impacts:
* has changed logging implementation (logback is now used behind slf4j instead of log4j)
* cannot easily take advantage of slf4j ability to switch implementations
* produces WARNING's during build and execution:

The following is an example of the WARNING
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/travis/.m2/repository/ch/qos/logback/logback-classic/1.1.2/logback-classic-1.1.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/travis/.m2/repository/org/slf4j/slf4j-log4j12/1.7.7/slf4j-log4j12-1.7.7.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]

crysmi:
As @gehel and @lookfirst have commented, the use of a pure slf4j approach to logging would be best as it would allow logging implementation details to be chosen at runtime.

This works for the logging aspect of jmxtrans but is harder to achieve for the output writers that use logging frameworks to produce user configurable file output.

Removing the concrete slf4j implementation classes from KeyOutWriter and TimeBasedRollingKeyOutWriter would appear to break/change the expected function of these output writers. SLF4J on its own does not appear to allow the level of runtime configuration that is currently required by these two classes.

SLF4J should be removed as a dependency of these two classes except to produce the standard getLogger logging. The output written by these two components should not go through SLF4J but instead use only direct log4j or logback-core classes

gehel:
`TimeBasedRollingKeyOutWriter` has a few direct dependency to logback-classic. As logback-classic alsoprovides the slf4j binding, we end up with 2 bindings on the classpath. It seems that it should be possible with not too much effort to remove this dependency, possibly by writing a few custom classes. We should be able to keep the dependency on logback-core.

Problematic dependencies are:

* `ch.qos.logback.classic.Level`
[...]

PR #224, opened 2015-02-12
Fix broken logging 

praste:
This fixes issue #221 . 
We have both log4j and logback jars and looks like slf4j is hitting logback jar first(as classloader seems to be loading those before log4j). In this case logback ignore log4j.xml and uses default configuration where it spits out everything to console. Default console logger is synchronous and hence it could be performance issue as well if we are logging too much.

Logback has really cool features like ability to change log level at runtime using JMX (which is available in log4j2 but not in log4j)

I have put together logback.xml to mimic intending log configuration we were trying to achieve using log4j.xml, with some more added features like automatically delete log file more than 15 days old.


gehel:
This is a rework of #223. As I said there, this strictly breaks compatibility by changing log4j.xml to logback.xml.

I think it still make sense to merge this as if fixes the long standing issue of logging being broken in JmxTrans. That being said, with my too fast merge on the first PR, I'm going to wait for @lookfirst opinion this time ;-)

lookfirst:
:+1:

praste:
I am not sure why my fix is not showing up in the master branch. I still
see a log4j.xml file in 'src' folder and no 'logback.xml there.

gehel:
Most probably my fault ... It was late yesterday, my rewrite was probably faulty. Apologies... Any chance you could resend another PR ?

praste:
Looks like I shot my self in the foot as well, now pull request says,
'there isn't anything to compare'

gehel:
I love git, except when I'm tired ...

praste:
where did logback go. I just pulled latest code and don't see logback.xml. 

Looks like someone inadvertently changed logging back you use log4j ?

gehel:
That's probably a mistake. In anycase, logging is mostly broken. We really need to remove all direct dependencies to log4j and logback (see #215). But this will probably require breaking backward compatibility.

Feel free to send a PR to add logback.xml again if you can tet that it works correctly. Integration test is of course more than welcomed!

PR #225, opened 2015-02-12
Fix of datagram socket factory

gehel:
This is a rebase of #77 (thanks to @Savar and sorry for the delay).

I should write an integration test for the DatagramSocketFactory, but it is fairly clear that the current implementation is broken.


lookfirst:
:+1:

PR #226, opened 2015-02-13
Fix logging issues

praste:
This fixes, closes issue #221 . 
We have both log4j and logback jars and looks like slf4j is hitting logback jar first(as classloader seems to be loading those before log4j). In this case logback ignore log4j.xml and uses default configuration where it spits out everything to console. Default console logger is synchronous and hence it could be performance issue as well if we are logging too much.

Logback has really cool features like ability to change log level at runtime using JMX (which is available in log4j2 but not in log4j)

I have put together logback.xml to mimic intending log configuration we were trying to achieve using log4j.xml, with some more added features like automatically delete log file more than 15 days old.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: java
-install: mvn compile -DskipTests=true -Dmaven.javadoc.skip=true -Dcobertura.skip=true -B -V
-before_script: "git clone -b travis `git config --get remote.origin.url` target/travis"
-script: mvn verify -B -V -PwithMutationTests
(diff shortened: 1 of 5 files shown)


gehel:
Thanks ! And sorry for all the troubles ...

jalaziz:
It appears something really odd is going on (or I'm terribly confused). I don't see logback.xml in the current HEAD, only log4j.xml. I see the revert from the first time, but wasn't this supposed to add it back?

PR #299, opened 2015-05-29
Pr scmsite

crysmi:
This pull request switches from using the github site plugin to maven-scm-publish-plugin as there was no current way to upload the whole site without triggering the abuse mechanisms

The scm-publish strategy relies on git scm provider which also has a problem for large changes as it tries to "git add" all changed files in one command line and this is way too big. 

The end result is to use site-stage to push each individual projects site to one "global" directory (currently found using session.executionRootDirectory) and scm-publish immediately after each site is staged. This minimises the number of files changed on each operation. A final site stage and scm-publish is called in order to pickup any aggregate reports that would only have been completed at the end of the site build (cobertura aggregate)

As I was interested in how the cobertura coverage was change (and why), Ive made the site upload to a folder called ${project.version} and disabled any deletion of file through scm-publish. This should mean you can maintain a site at gh-pages without the maven site overwriting it.

The use of the git provider, means authentication is still done with the github token but it is recommended to add the following environment variables to travis:
GIT_AUTHOR_NAME
GIT_COMMITTER_NAME
and these with secure settings applied so they dont show in the logs:
GIT_AUTHOR_EMAIL
GIT_COMMITTER_EMAIL

The email settings are used by github to determine what is displayed on github.

Git provider versions before 1.9.4 showed the whole token in the logs, so the version is set as a dependency with a fixed version

Ive added a required environment variable (SITE_BRANCH_NAME) that designates which branch should initiate the site build - this was a convenience method for me to test outside the main project. It should also mean no site will be built until you set this in travis to the appropriate branch

Ive also added a profile to turn on verbose logging on pitest (off by default)

There is an example of the complete build including site deploy with this branch at:
https://travis-ci.org/abcops/jmxtrans/builds/64401278
and the built site can be seen at:
http://abcops.github.io/jmxtrans/251-SNAPSHOT/

Im not sure what will be produced when a non-feature branch is built.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ install:
   - mvn compile -DskipTests=true -Dmaven.javadoc.skip=true -Dcobertura.skip=true -B -V
 script: ./travis-script.sh
 after_success:
-  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "$SITE_BRANCH_NAME" && mvn site --settings ${HOME}/travis/settings.xml -B -V && mvn site:stage scm-publish:publish-scm --settings ${HOME}/travis/settings.xml -B -V
+  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "$SITE_BRANCH_NAME" && mvn site-deploy --settings ${HOME}/travis/settings.xml -B -V && mvn site:stage scm-publish:publish-scm --settings ${HOME}/travis/settings.xml -B -V
 branches:
   except:
     - travis
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ install:
   - mvn compile -DskipTests=true -Dmaven.javadoc.skip=true -Dcobertura.skip=true -B -V
 script: ./travis-script.sh
 after_success:
-  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "master" && mvn site --settings ${HOME}/travis/settings.xml -B -V
+  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "$SITE_BRANCH_NAME" && mvn site --settings ${HOME}/travis/settings.xml -B -V && mvn site:stage scm-publish:publish-scm --settings ${HOME}/travis/settings.xml -B -V
 branches:
   except:
     - travis
--- a/jmxtrans-output/jmxtrans-output-core/pom.xml
+++ b/jmxtrans-output/jmxtrans-output-core/pom.xml
@@ -41,6 +41,10 @@
 			<artifactId>commons-lang</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>commons-logging</groupId>
+			<artifactId>commons-logging</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>commons-pool</groupId>
 			<artifactId>commons-pool</artifactId>
 		</dependency>
@@ -118,16 +122,6 @@
 
 	<build>
 		<plugins>
-			<plugin>
-				<groupId>com.github.github</groupId>
-				<artifactId>site-maven-plugin</artifactId>
-				<configuration>
-					<merge>true</merge>
-					<message>Creating site for ${project.name}
-						${project.version}</message>
-					<path>jmxtrans-output/${project.artifactId}</path>
-				</configuration>
-			</plugin>
 			<plugin>
 				<groupId>net.ju-n.maven.plugins</groupId>
 				<artifactId>checksum-maven-plugin</artifactId>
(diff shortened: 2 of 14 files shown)


gehel:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/jmxtrans/jmxtrans/pull/299'>Refresh</a>.<a href='#crh-update-none'></a>

gehel:
Sorry for not reacting before. This looks good! I need to find some quite time to setup the correct variables and merge this...

Thanks a lot!

jmxtrans-output/jmxtrans-output-core/pom.xml
```diff
@@ -41,6 +41,10 @@
 			<artifactId>commons-lang</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>commons-logging</groupId>
```
gehel: Why did you add this dependency to commons-logging? It seems that there is no code using it directly...

pom.xml
```diff
@@ -649,6 +652,44 @@
...
+						<!-- try to shortcut the delete process since we are not doing it anyway -->
+						<ignorePathsToDelete>
+							<ignorePathsToDelete>%regex[.*]</ignorePathsToDelete>
+						</ignorePathsToDelete>
+					</configuration>
+					<dependencies>
+						<dependency>
+							<groupId>org.apache.maven.scm</groupId>
+							<artifactId>maven-scm-api</artifactId>
+							<version>[1.9.4]</version>
```
gehel: Why using version range instead of fixed version number ? I'm not an expert, but it seems to me that `[1.9.4]` is equivalent to just `1.9.4`. Did I miss something ?

pom.xml
```diff
@@ -960,6 +1066,10 @@
 				<artifactId>versions-maven-plugin</artifactId>
 				<version>2.1</version>
 			</plugin>
+			<plugin>
```
gehel: pitest now has a report goal ! Great !

:+1:

pom.xml
```diff
@@ -714,23 +780,25 @@
...
+							<haltOnFailure>${verify.haltOnFailure}</haltOnFailure>
 							<lineRate>0</lineRate>
 							<packageBranchRate>0</packageBranchRate>
 							<packageLineRate>0</packageLineRate>
 							<totalBranchRate>${verify.totalBranchRate}</totalBranchRate>
 							<totalLineRate>${verify.totalLineRate}</totalLineRate>
 						</check>
+						<instrumentation>
+							<excludes>
+								<exclude>**/Test*.java</exclude>
```
gehel: Why do we need those excludes ? There might be a good reason, in which case an inline comment would be welcomed.

.travis.yml
```diff
@@ -6,7 +6,7 @@ install:
   - mvn compile -DskipTests=true -Dmaven.javadoc.skip=true -Dcobertura.skip=true -B -V
 script: ./travis-script.sh
 after_success:
-  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "master" && mvn site --settings ${HOME}/travis/settings.xml -B -V
+  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "$SITE_BRANCH_NAME" && mvn site-deploy --settings ${HOME}/travis/settings.xml -B -V && mvn site:stage scm-publish:publish-scm --settings ${HOME}/travis/settings.xml -B -V
```
gehel: Not sure what this `$SITE_BRANCH_NAME` variable is. I would expect to publish maven site only for master, as feature branches typically contain code that is in progress, so should probably not be published (we do not publish snapshots to Sonatype either). I also do not see where this variable is set.

travis-script.sh
```diff
@@ -40,7 +40,7 @@ if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
     fi
   else
     echo "Building feature branch"
-    mvn verify --settings ${MVN_SETTINGS} -B -V -PwithMutationTests,rpm,deb
+    mvn install --settings ${MVN_SETTINGS} -B -V -PwithMutationTests,rpm,deb
```
gehel: Why do you need to install the artifacts from a feature branch ? Is it related to the `$SITE_BRANCH_NAME` variable used in `.travis.yml` ? Is there a requirement to have the artifacts deployed locally to create the maven site ? If that's the case, we need to decide if we want to publish maven site for feature branches.

gehel:
Thanks a lot ! This looks good ! The aggregated reports is also something I wanted to do for a long time, but never took the time to actually do it.

I still have a few questions (mainly because I'm not smart enough to understand all you did), but big :+1: !

gehel:
General question: do you think it is a good idea to extract `GIT_AUTHOR_NAME` and other such variables from the last git commit (which should identify the person who merged the code) or do you think we should configure a static user / committer ?

gehel:
@zonArt If you still have too much time at night, this PR would welcome some of your love...

PR #328, opened 2015-09-04
Fixed Suse groupadd/useradd

magullo:
I have removed fixed guid/uid number, letting system decide.
A problem could happend though, during an upgrade: already created folder could not be updated, if guid/uid is different from the previous one.
The %pre script still checks for RHEL5, otherwise uses default script for everything else.
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/jmxtrans/jmxtrans/pull/328?approve=1'></a>&nbsp;<a href='https://github.com/jmxtrans/jmxtrans/pull/328'></a>
<a href='#crh-end'></a>

### DIFF
--- a/jmxtrans/pom.xml
+++ b/jmxtrans/pom.xml
@@ -322,18 +322,15 @@
 							</defineStatements>
 							<preinstallScriptlet>
 								<script>if [ $1 = 1 ]; then
-									USER_ID=${dollar.sign}(id -u ${package.user} 2&gt;/dev/null)
-									if [ -z "$USER_ID" ]; then
-										REDHAT_RELEASE=${dollar.sign}(lsb_release -sr|cut -d"." -f1)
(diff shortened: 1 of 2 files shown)


gehel:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/jmxtrans/jmxtrans/pull/328'>Refresh</a>.<a href='#crh-update-none'></a>

gehel:
Looks good. I'm merging this so we can have some feedback from users using the SNAPSHOT version. I'll wait a bit before doing a release ...

I'm still not convinced we have the right solution for managing the UID ... but I do not have a bettre proposal ...

magullo:
Yeah I know, I will keep on looking for other solutions around, if I find one which better suits jmxtrans.
Thanks

PR #405, opened 2016-01-14
Deduplicate output writers

### LINKED ISSUE #396: Deduplicate OutputWriters
At the moment, `OutputWriter`s are instantiated for each query. Which means that the same `OutputWriter` exist multiple times. One way of fixing this would be to change the configuration format and have queries reference `OutputWriter`s. Not very keen on this idea as it means a non backward compatible change to configuration format. Another way would be to merge `OutputWriter`s during configuration parsing. This should not be too hard to do, with minimal changes to [ConfigurationParser](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/ConfigurationParser.java).

gehel:
Fixes #396


gehel:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/jmxtrans/jmxtrans/pull/405'>Refresh</a>.<a href='#crh-update-none'></a>

PR #427, opened 2016-03-08
Update ElasticWriter.java

lmx1989219:
only set timestamp to type date the es will work well in use es dynamic mapping

### DIFF
--- a/jmxtrans-output/jmxtrans-output-elastic/src/main/java/com/googlecode/jmxtrans/model/output/elastic/ElasticWriter.java
+++ b/jmxtrans-output/jmxtrans-output-elastic/src/main/java/com/googlecode/jmxtrans/model/output/elastic/ElasticWriter.java
@@ -128,7 +128,7 @@ public class ElasticWriter extends BaseOutputWriter {
 					map.put("key", values.getKey());
 					map.put("keyAlias", result.getKeyAlias());
 					map.put("value", Double.parseDouble(value.toString()));
-					map.put("timestamp", result.getEpoch());
+					map.put("timestamp", java.util.Date(result.getEpoch()));
 
 					log.debug("Insert into Elastic: Index: [{}] Type: [{}] Map: [{}]", indexName, ELASTIC_TYPE_NAME, map);
 					Index index = new Index.Builder(map).index(indexName).type(ELASTIC_TYPE_NAME).build();


gehel:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/jmxtrans/jmxtrans/pull/427'>Refresh</a>.<a href='#crh-update-none'></a>

jmxtrans-output/jmxtrans-output-elastic/src/main/java/com/googlecode/jmxtrans/model/output/elastic/ElasticWriter.java
```diff
@@ -128,7 +128,7 @@ protected void internalWrite(Server server, Query query, ImmutableList<Result> r
 					map.put("key", values.getKey());
 					map.put("keyAlias", result.getKeyAlias());
 					map.put("value", Double.parseDouble(value.toString()));
-					map.put("timestamp", result.getEpoch());
+					map.put("timestamp", java.util.Date(result.getEpoch()));
```
gehel: Could you `import java.util.Date` ?

gehel:
This is a backward incompatible change, so I'm not too keen on merging it. I would prefer fixing the mapping by adding an explicit format.

Side note: we need some integration tests with elasticsearch.

jmxtrans-output/jmxtrans-output-elastic/src/main/java/com/googlecode/jmxtrans/model/output/elastic/ElasticWriter.java
lmx1989219: image
this is ok

gehel: This is still changing the format of an existing field. If we have users already relying on the current behavior (and we probably have) it will break things for them. Adding a new field, while ugly, is the right way to go.

Side note: could you paste inline code instead of screenshots? That makes it much easier to search...

lmx1989219: i'm so sorry.thanks for your suggestion.

gehel: @lmx1989219: No problem! You are trying to fix something that is broken, I applaud your efforts! It is non trivial to fix it without breaking it for users who are relying on the current behavior, but that's the fun part!

gehel:
I actually took a few minutes to read the [`ElasticWriter`](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-elastic/src/main/java/com/googlecode/jmxtrans/model/output/elastic/ElasticWriter.java) code. We do [create a mapping at startup](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-elastic/src/main/java/com/googlecode/jmxtrans/model/output/elastic/ElasticWriter.java#L176). The [mapping](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-elastic/src/main/resources/elastic-mapping.json#L43) is probably just missing the [appropriate format](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#built-in-date-formats).

I would need more time than I have right now to actually test and validate any change.

@lmx1989219 you should be able to modify the mapping in your elasticsearch instance (without touching JmxTrans code at all) and validate if that works or not...

lmx1989219:
that's so cool.

lmx1989219:
hi gehel,
i found a problem that the index is declared by construct method and the es data folder can not empty with date.i suggest we can get client dynamically so that es can management the index data well 
 according to date.

e.g:
test_jmx-entries-2016.03.12
test_jmx-entries-2016.03.13
test_jmx-entries-2016.03.14
....

gehel:
@lmx1989219 I'm not sure I understand what you mean... If you have working code, please send it along as the basis for the discussion...

lmx1989219:
@gehel 
`@JsonCreator
	public ElasticWriter(@JsonProperty("typeNames") ImmutableList<String> typeNames,
			@JsonProperty("booleanAsNumber") boolean booleanAsNumber, @JsonProperty("rootPrefix") String rootPrefix,
			@JsonProperty("debug") Boolean debugEnabled, @JsonProperty("connectionUrl") String connectionUrl,
			@JsonProperty("settings") Map<String, Object> settings) throws IOException {

		super(typeNames, booleanAsNumber, debugEnabled, settings);

		this.rootPrefix = firstNonNull(rootPrefix, (String) getSettings().get("rootPrefix"), DEFAULT_ROOT_PREFIX);

		this.connectionUrl = connectionUrl;
		# this.indexName = this.rootPrefix + "_jmx-entries"+"yyyy.MM.dd";
		# this.jestClient = createJestClient(connectionUrl);
	}
`

**the '#' row copy to follow code**


`protected void internalWrite(Server server, Query query, ImmutableList<Result> results) throws Exception {
               //**the '#' row insert here**
		for (Result result : results) {
			log.debug("Query result: [{}]", result);
			Map<String, Object> resultValues = result.getValues();
			for (Entry<String, Object> values : resultValues.entrySet()) {
				Object value = values.getValue();
				if (isNumeric(value)) {

					Map<String, Object> map = new HashMap<String, Object>();
					map.put("serverAlias", server.getAlias());
					map.put("server", server.getHost());
					map.put("port", server.getPort());
					map.put("objDomain", result.getObjDomain());
					map.put("className", result.getClassName());
					map.put("typeName", result.getTypeName());
					map.put("attributeName", result.getAttributeName());
					map.put("key", values.getKey());
					map.put("keyAlias", result.getKeyAlias());
					map.put("value", Double.parseDouble(value.toString()));
					 map.put("timestamp", result.getEpoch());
//					map.put("timestamp", sdf.format(new Date(result.getEpoch())));
					log.debug("Insert into Elastic: Index: [{}] Type: [{}] Map: [{}]", indexName, ELASTIC_TYPE_NAME,
							map);
					Index index = new Index.Builder(map).index(indexName).type(ELASTIC_TYPE_NAME).build();
					JestResult addToIndex = jestClient.execute(index);
					if (!addToIndex.isSucceeded()) {
						throw new ElasticWriterException(
								String.format("Unable to write entry to elastic: %s", addToIndex.getErrorMessage()));
					}
				} else {
					log.warn("Unable to submit non-numeric value to Elastic: [{}] from result [{}]", value, result);
				}
			}
		}
	}`

gehel:
Wow! A pull request would be easier to read...

PR #429, opened 2016-03-15
Update ElasticWriter.java

lmx1989219:
my suggestion like this code


lmx1989219:
@gehel 
goodmorning gehel ,
this code is my meaning.

gehel:
Ok, I see what you mean. To have a daily rolling index, we also need to configure an [index template](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html), so that new indexes have the correct configuration. I'm writing a new (and hopefully improved) ElasticWriter to include those changes without breaking things for people who already use the current one. You can have a look at the [branch](https://github.com/jmxtrans/jmxtrans/tree/improve-elasticsearch-writer) (work in progress, might not work or even compile).

PR #470, opened 2016-06-08
Close invalidated JMX Connection when query execution fails.

vveloso:
When an error occurs during execution of a JMX query, the JMXConnection object is invalidated in the pool. This means that future queries will not reuse the existing JMXConnection, rather creating a new one.

We should ensure that the JMXConnection is closed before invalidating/discarding it in order to avoid resource leaks.

There is no knowledge about the original query error, which means that the JMXConnection may be in an invalid state. Hence just ignore and log any exceptions that may be thrown when we try to close it on error.
<a href='#crh-start'></a><a href='#crh-data-%7B%7D'></a>
<a href='https://www.codereviewhub.com/'></a>
<a href='https://www.codereviewhub.com/jmxtrans/jmxtrans/pull/470?approve=1'></a>&nbsp;<a href='https://github.com/jmxtrans/jmxtrans/pull/470'></a>
<a href='#crh-end'></a>

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
@@ -24,9 +24,9 @@ package com.googlecode.jmxtrans.model;
 
 import com.fasterxml.jackson.annotation.JacksonInject;
 import com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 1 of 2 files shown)


gehel:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/jmxtrans/jmxtrans/pull/470'>Refresh</a>.<a href='#crh-update-none'></a>

gehel:
Thanks for the fix!

Issue #500, opened 2016-10-22
ResultTransformerOutputWriter - Not stopped/cleaned up


willmpls:
I'm working on a UDP/TCP output for InfluxDB line protocol, and noticed that when using the ResultTransformerOutputWriter along with a single output defined at the server level I accumulate open TCP or UDP ports as I add or update new config files faster than I should.  

Root cause is that the writers aren't being cleaned up, so open ports continue to pile up until jmxtrans is restarted.

This behavior appears to be due to the following:
- JMXTransformer only loops and calls stop() on writers defined in queries, not servers.
- ResultTransformerOutputWriter doesn't implement the stop() method 

The following diff shows updates I made which fix the issue in my case, but I wanted to get thoughts on implementation before submitting a pr:

https://github.com/jmxtrans/jmxtrans/compare/master...willmpls:output_writer_stop

gehel:
This is definitely a bug. [OutputWriter](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/OutputWriter.java) should extend [AutoCloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) instead of declaring a specific `close()` method, and obviously, all output writers should take care of cleaning their resources.

Thanks for spotting that!

gehel:
Your proposed change looks good. Minor comment: I would let the exception bubble up in [`ResultTransformerOutputWriter.close`](https://github.com/jmxtrans/jmxtrans/compare/master...willmpls:output_writer_stop#diff-5e0b720cd7f46e983925f01c06b45470R77) instead of logging it.

gehel:
Closing this as it should be fixed by #503

PR #503, opened 2016-10-24
Singleton OutputWriter cleanup, implement stop() on ResultTransformerOutputWriter

willmpls:
This PR addresses #500 to fix cleanup of singleton output writers along with implementing a stop() method on ResultTransformerOutputWriter.  See issue #500 for additional details.

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/output/support/ResultTransformerOutputWriter.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/output/support/ResultTransformerOutputWriter.java
@@ -31,6 +31,8 @@ import com.googlecode.jmxtrans.model.results.BooleanAsNumberValueTransformer;
 import com.googlecode.jmxtrans.model.results.IdentityValueTransformer;
 import com.googlecode.jmxtrans.model.results.ResultValuesTransformer;
 
+import com.googlecode.jmxtrans.exceptions.LifecycleException;
+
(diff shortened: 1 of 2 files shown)


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/output/support/ResultTransformerOutputWriter.java
```diff
@@ -66,5 +68,12 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
 		return new ResultTransformerOutputWriter<>(new ResultValuesTransformer(new IdentityValueTransformer()), target);
 	}
 
+	public void stop() throws LifecycleException {
+		try {
+			target.stop();
+		} catch (LifecycleException ex) {
```
willmpls: Agreed, fixed!

gehel:
There is actually a few other OutputWriters that have the same issue. I have not gone through the whole list but at least HttpOutputWriter and WriterPoolOutputWriter have the same issue. If you have time to fix them as well, that would be great! Else I'll have a look as soon as I can.

willmpls:
I can definitely help out with some of them.  Would you be able to generate the definitive list of ones that need fixing?  Just want to make sure I don't miss any.

gehel:
I can go through the OutputWriter hierarchy and check, but at this point it is not much more work to fix them directly. I'll merge this PR already and I'll have a look at the whole hierarchy asap.

PR #504, opened 2016-10-24
Create KeyOutOverWriter

bainss:
Allow output of jmx stats at timely intervals, overwriting the original


gehel:
Those log4j based output writers are problematic in quite a few ways. The main issue is that they introduce a direct dependency on log4j and on log4j internals, causing all kind of trouble. The same kind of behaviour could be achieved by using `Slf4JOutputWriter` and a logger configuration that outputs in with whatever appender you want.

PR #505, opened 2016-10-24
InfluxDB TCP/UDP Line Protocol support

willmpls:
Based extensively off TCollector UDP writer.  Adds "allKeysAsTags" parameter to process all mbean key/values as tag pairs.

Comments welcome!

### DIFF
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDBTCPWriter2.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDBTCPWriter2.java
@@ -1,0 +1,54 @@
+/**
+ * The MIT License
+ * Copyright © 2010 JmxTrans team
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.googlecode.jmxtrans.model.output;
+
+import com.googlecode.jmxtrans.model.Query;
+import com.googlecode.jmxtrans.model.Result;
+import com.googlecode.jmxtrans.model.Server;
+import com.googlecode.jmxtrans.model.output.support.WriterBasedOutputWriter;
+import com.googlecode.jmxtrans.model.output.support.influxdb.InfluxDBMessageFormatter;
+
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.ThreadSafe;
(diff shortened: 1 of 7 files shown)


willmpls:
I'll clean the test errors up.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDBTCPWriter2.java
```diff
@@ -0,0 +1,54 @@
...
+import com.googlecode.jmxtrans.model.output.support.WriterBasedOutputWriter;
+import com.googlecode.jmxtrans.model.output.support.influxdb.InfluxDBMessageFormatter;
+
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.ThreadSafe;
+import java.io.IOException;
+import java.io.Writer;
+
+@ThreadSafe
+public class InfluxDBTCPWriter2 implements WriterBasedOutputWriter {
```
gehel: It looks like the TCP and UDP writers are exactly the same. This is as it should, they don't need to actually know anything about the protocol, they just see a writer. They could probably be merged in a single `InfluxDBWriter` (or similar name) and have different factories building them.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/influxdb/InfluxDBMessageFormatter.java
```diff
@@ -0,0 +1,278 @@
...
+									@Nonnull ImmutableMap<String, String> tags,
+									@Nonnull String tagName,
+									@Nullable String metricNamingExpression,
+									boolean mergeTypeNamesTags,
+									boolean allKeysAsTags,
+									@Nullable String hostnameTag) throws UnknownHostException, LifecycleException {
+		this.typeNames = typeNames;
+		this.tags = tags;
+		this.tagName = tagName;
+		if (metricNamingExpression != null) {
```
gehel: The construction of the metricNameStrategy should probably be done in the Factory. Let's try to keep constructors as simple as possible.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDBTCPWriter2Test.java
```diff
@@ -0,0 +1,64 @@
...
+  private InfluxDBTCPWriter2 writer;
+  private Writer outputWriter;
+  private Result result;
+
+  @Before
+  public void setup() {
+    influxDBMessageFormatter = Mockito.mock(InfluxDBMessageFormatter.class);
+    writer = new InfluxDBTCPWriter2(influxDBMessageFormatter);
+    outputWriter = Mockito.mock(Writer.class);
+    result = Mockito.mock(Result.class);
```
gehel: Mocking `Result`s is actually not necessary here. You could just use the `ResultFixtures` class.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDBTCPWriter2Test.java
```diff
@@ -0,0 +1,64 @@
...
+  private InfluxDBMessageFormatter influxDBMessageFormatter;
+  private InfluxDBTCPWriter2 writer;
+  private Writer outputWriter;
+  private Result result;
+
+  @Before
+  public void setup() {
+    influxDBMessageFormatter = Mockito.mock(InfluxDBMessageFormatter.class);
+    writer = new InfluxDBTCPWriter2(influxDBMessageFormatter);
+    outputWriter = Mockito.mock(Writer.class);
```
gehel: No need to mock `Writer`. Using a `StringWriter` would work (and be simpler IMHO). If we can test without Mockito, we should!

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDBTCPWriterFactory.java
```diff
@@ -0,0 +1,102 @@
...
+import java.net.UnknownHostException;
+import java.util.Map;
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.googlecode.jmxtrans.model.output.support.pool.FlushStrategyUtils.createFlushStrategy;
+
+@ThreadSafe
+@EqualsAndHashCode
+@ToString
+public class InfluxDBTCPWriterFactory implements OutputWriterFactory {
```
gehel: We could combine TCP and UDP writer factories in a single factory, which take a `protocol` param which can be TCP or UDP. That would remove a lot of duplication!

gehel:
I put some of my ideas in code. This might be clearer than my comments on this PR. There are still issues with that code, but it might help you anyway. Have a look at https://github.com/jmxtrans/jmxtrans/tree/influxdb_lineproto

PR #507, opened 2016-11-01
Fix encoding declaration in Python code

beevee:
As of https://www.python.org/dev/peps/pep-0263/, magic comment must be the first or the second line of the file. Also, copyright symbol in license is UTF-8.

```
File "/usr/share/jmxtrans/tools/yaml2jmxtrans.py", line 4
SyntaxError: Non-ASCII character '\xc2' in file /usr/share/jmxtrans/tools/yaml2jmxtrans.py on line 4, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```


gehel:
Looks good, thanks!

PR #508, opened 2016-11-02
Add Docker image

zepouet:
Purpose a docker image for JMX Trans.

Two commands are available : start and jmx.
I use gosu and tini to reap zombies and we can choose the build release with arguments.

I hope it is nice and hope it is right located into the sources.

Thanks


gehel:
That looks great! I have a few comments, but I need some time to put them together. Thanks a lot!

hgomez:
"This image is officially deprecated in favor of the openjdk image" ... Why not moving to openjdk (jre) right now ?

zepouet:
Hello @gehel @hgomez 

I modified the code for the openjdk. Thanks for all.

Best regards,
Nicolas

jmxtrans/docker/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+ENV CONTINUE_ON_ERROR false
+ENV LOG_DIR /var/log/jmxtrans
+ENV JSON_DIR /var/lib/jmxtrans
+ENV GC_OPTS=${GC_OPTS:-"-Xms${HEAP_SIZE}m -Xmx${HEAP_SIZE}m"}
+ENV JMXTRANS_OPTS="$JMXTRANS_OPTS -Djmxtrans.log.level=${LOG_LEVEL} -Djmxtrans.log.dir=${LOG_DIR}"
+ENV MONITOR_OPTS=" -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=2101  -Dcom.sun.management.jmxremote.rmi.port=2101 "
+
+# Install jmxtrans
+RUN echo "debconf debconf/frontend select noninteractive" | debconf-set-selections
+RUN wget http://central.maven.org/maven2/org/jmxtrans/jmxtrans/${JMXTRANS_VERSION}/jmxtrans-${JMXTRANS_VERSION}.deb
```
gehel: It would be nice to install the jmxtrans version available locally instead of downloading it from Maven central. This would ensure that it is up to date and aligned with the local sources. That would be a step on the path of building (and publishing) Docker images from the standard jmxtrans build.

jmxtrans/docker/Dockerfile
```diff
@@ -0,0 +1,70 @@
+FROM openjdk:8u102-jdk
```
gehel: Missing license header (note that `mvn license:format` should add those headers automatically for you)

jmxtrans/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
gehel: Missing license header

jmxtrans/docker/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+docker run -d -v `pwd`/json-files:/var/lib/jmxtrans jmxtrans/jmxtrans 
+docker run -d -v `pwd`/json-files:/var/lib/jmxtrans jmxtrans/jmxtrans start 
+docker run -d -v `pwd`/json-files:/var/lib/jmxtrans -P jmxtrans/jmxtrans jmx
+docker run -d -v `pwd`/json-files:/var/lib/jmxtrans -p 9999:2101 jmxtrans/jmxtrans jmx
+```
+
+You have two commands available :
+* start (default value)  
+* jmx to start the jvm with jmx debug mode
```
gehel: I'm not convinced by the naming. Having a `start` and a `jmx` commands let me think that `jmx` will not start jmxtrans (when actually it does start it). Also, I'm not sure that there is really a reason to not publish JMX. Extra runtime cost is negligible and it is not exposed anyway. It might make sense to always enable JMX and to provide a default config that makes jmxtrans monitor itself.

gehel:
Thanks for this docker image! That is a good step forward!

It might make sense to move this to a specific maven module and have the docker image be built during packaging phase (and ideally uploaded to Docker hub).

jmxtrans/docker/Dockerfile
zepouet: Thanks for reviews.

Maybe I am wrong but it is easier to build image into dockerhub.
Jenkins and others does the same process : https://github.com/jenkinsci/docker/blob/2fb0684fe6eb11a3c1145d71e0cb50bf2dda8d11/Dockerfile

Therefore I could use .jar rather than .deb but with .deb you create jmxtrans user.

Another solution is to checkout the source into the Dockerfile and run the build processus.
If you prefer, I can do it !

jmxtrans/docker/README.md
zepouet: I am not a jmx expert. I wonder about the cost if it is always enabled.
So you are right we could always enable it.
I modify it.

jmxtrans/docker/docker-entrypoint.sh
zepouet: Done. Thanks !

jmxtrans/docker/Dockerfile
zepouet: Done. Thanks !

jmxtrans/docker/README.md
zepouet: I enabled it by default and keep an option to disable it.
Documentation is updated.

Best regards,

jmxtrans/docker/Dockerfile
gehel: I'm fairly new to Docker, so my comments might not make sense. Still, I'll try...

The jmxtrans user is not really needed, my understanding is that in a dock, running jmxtrans as the default (root) user is perfectly fine.

It would probably make sense to have logging directed only to stdout (not to file) to keep writes inside the dock to a minimum.

So we don't actually use anything from the .deb package, except the .jar file. It might make more sense to not use the .deb and just download the .jar.

Keep in mind that I have a fairly limited understanding of docker, the above comments might be complete crap...

zepouet: Your comments are right. In fact Perfect is the enemy of Good.
You can always do better with Dockerfile : so why use gosu, tini... and dedicated user ?
I just try to have a good average between complexity and power.
It is nice to have a dedicated user for mapping rights between host and container (namespace). For security too. Not mandatory but better. I can create easily it without .deb.

For logs you are totally right. I added to soft link to stdout at the end of the Dockerfile for this reason. It was a nice workaround. 
The perfect solution would be to have a maven profil with a dedicated configuration for logs or jar parameters. So I build source release into dockerfile, i could choose it.
Logs are an important point for docker. Only logs from pid 1 are streamed. (like only pid 1 receives stop signal) 

I will purpose to you another image very soon with jar and the jmxtrans user if you are right.
Comments are welcome.

Thanks

gehel: For logs, no need to create a specific jar. The log4j configuration can be overwritten from outside of the jar, by adding a log4j.xml file to the classpath, or by setting `-Dlog4j.configuration=...`.

If I'm bikeshedding, let me know, I'll merge whatever you find reasonable. We can always improve later if needed...

gehel: More by curiosity and so that I might learn something: what is the use of tini in this context? jmxtrans runs as a single Java process and mostly never forks subprocesses (well, `RRDToolWriter` does fork subprocesses, some other output writers might as well, but that's an edge case and in most cases those should not be used). I'm wondering...

hgomez: Tini seems to be a de factor signal handling in Docker envs, ie https://github.com/krallin/tini/issues/8

gehel: I'm learning things. And yes, it does seem that tini makes some kind of sense (probably quite limited in this specific case, but why not).

Thanks @hgomez for the link!

zepouet: Sorry for my being late but Henri was right so I took a little time :)
Yes tini is nice to reap zombies and to be sure that signal are right transmitted.

I replaced .deb with .jar and use now a log4j.xml as you told me. It is a nice workaround.

But I have a little problem. I cannot change the log level at docker runtime with
`docker run ... --env LOG_LEVEL=warn`

```
    <root>
        <level value="$${env:LOG_LEVEL}" />
        <appender-ref ref="console" />
    </root>
```

It is strange... the other solution is to mount at runtime the right log4j.xml file.

```
docker run -d 
 -v `pwd`/log4j-warn.xml:/usr/share/jmxtrans/conf/log4j.xml  
 -v `pwd`/json-files:/var/lib/jmxtrans cloudunit/jmxtrans
```
I think it is in fact the best solution. So an user can choose it own custom file.

Thanks for all ! Nicolas

jmxtrans/docker/log4j.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
gehel: Missing license header.

jmxtrans/docker/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+	&& wget -O /usr/local/bin/tini.asc "https://github.com/krallin/tini/releases/download/$TINI_VERSION/tini.asc" \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& gpg --keyserver ha.pool.sks-keyservers.net --recv-keys 6380DC428747F6C393FEACA59A84159D7001A4E5 \
+	&& gpg --batch --verify /usr/local/bin/tini.asc /usr/local/bin/tini \
+	&& rm -r "$GNUPGHOME" /usr/local/bin/tini.asc \
+	&& chmod +x /usr/local/bin/tini \
+	&& tini -h
+
+ARG JMXTRANS_VERSION
+ENV JMXTRANS_VERSION ${JMXTRANS_VERSION:-259}
```
gehel: It would be really nice to get the latest release! [`mvn dependency:get`](http://maven.apache.org/plugins/maven-dependency-plugin/get-mojo.html) could do that automagically, but installing maven for such a simple use case is probably overkill. Anyway, this is just a nice to have at the moment...

gehel:
Let's already merge this, it looks good enough for now. There are still a few things I would like to see, but this can come in a second phase:

* version number should come from pom.xml and not be duplicated
* all the code should be pushed to a specific module
* it would be nice to have the docker image built and pushed to docker hub on each release, or even to publish SNAPSHOT.

zepouet:
Thanks ! I just wanted to leave a comment after my last commit.
I'm going to see about the last points.

gehel:
I just updated the conifg on https://hub.docker.com/r/jmxtrans/jmxtrans/, let's see if it works. @zepouet thanks a lot for all the work!

gehel:
another stupid question: Is it necessary to have a specific version of openjdk as the base image? Or should we just use openjdk:8 ?

zepouet:
No stupid question :) You can use major release rather than minor value.
You can change it

Issue #484, opened 2016-08-03
YAML outputWriter config for GraphiteWriterFactory requires "settings" key


vemod:
1. When using `outputWriters` configuration in YAML `settings` wrapper key is required. Thus produced JSON config for GraphiteWriterFactory outputWriter is created wrong. 
   `com.googlecode.jmxtrans.model.output.GraphiteWriterFactory` doens't support `settings: {}` wrapper
2. Documentation is also outdated and doesn't provide required information about outputWriters configuration.
## 

At the moment 

```
outputWriters: 
    - "@class": "com.googlecode.jmxtrans.model.output.GraphiteWriterFactory"
      settings:
        host: "graphite"
        port: "2003"
        flushStrategy: "always"
        rootPrefix: "servers"
```

produces

```
"outputWriters": [
      {
       "@class": "com.googlecode.jmxtrans.model.output.GraphiteWriterFactory",
       "settings": {
        "flushStrategy": "always",
        "typeNames": [
         "name"
        ],
        "host": "graphite",
        "port": "2003",
        "rootPrefix": "servers"
       }
      }
```

BUT `com.googlecode.jmxtrans.model.output.GraphiteWriterFactory` expects the configuration without `settings: {}` wrapper. 

So when `settings:` wrapper is skipped following exception is thrown

```
Traceback (most recent call last):
  File "../yaml2jmxtrans.py", line 280, in <module>
    servers = q.create_host_set_configuration(s['hosts'],s['query_names'], query_port, s['username'], s['password'], s['urlTemplate'], set_name, global_host_alias)
  File "../yaml2jmxtrans.py", line 173, in create_host_set_configuration
    root['servers'].append(self.create_host_entry(host, query_names, port, username, password, urlTemplate, alias, set_name, rootPrefix))
  File "../yaml2jmxtrans.py", line 126, in create_host_entry
    hostentry['queries'].append(self.create_query_entry(query_name, rootPrefix))
  File "../yaml2jmxtrans.py", line 114, in create_query_entry
    queryentry['outputWriters'] = self.create_output_writer_configuration(typeName, rootPrefix)
  File "../yaml2jmxtrans.py", line 200, in create_output_writer_configuration
    writer[iter]['settings']['typeNames'] = typeNames
TypeError: 'NoneType' object does not support item assignment
```

Expected YAML: 

```
outputWriters: 
    - "@class": "com.googlecode.jmxtrans.model.output.GraphiteWriterFactory"
[...]

PR #509, opened 2016-11-02
fix issue #484

demonix:
yaml2jmxtrans doesn't write settings element to json config now


gehel:
I need some time to review this properly. And we don't have any tests for this... that's bad... I'll see if I can do something about it.

Issue #489, opened 2016-09-19
"error parsing query" while writing to InfluxDB


daniccan:
I am getting the following error while trying to write to InfluxDB using jmxtrans.

java.lang.RuntimeException: {"error":"error parsing query: found NOT, expected ; at line 1, char 20"}

```
at org.influxdb.impl.InfluxDBErrorHandler.handleError(InfluxDBErrorHandler.java:19)
at retrofit.RestAdapter$RestHandler.invoke(RestAdapter.java:242)
at org.influxdb.impl.$Proxy27.query(Unknown Source)
at org.influxdb.impl.InfluxDBImpl.createDatabase(InfluxDBImpl.java:205)
at com.googlecode.jmxtrans.model.output.InfluxDbWriter.doWrite(InfluxDbWriter.java:151)
at com.googlecode.jmxtrans.model.output.support.ResultTransformerOutputWriter.doWrite(ResultTransformerOutputWriter.java:50)
at com.googlecode.jmxtrans.jmx.ResultProcessor$1.run(ResultProcessor.java:58)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
at java.util.concurrent.FutureTask.run(FutureTask.java:262)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
at java.lang.Thread.run(Thread.java:744)
```

My InfluxDB version is 1.0.0. 

Following is the example.json I am using,

{
  "servers" : [ {
    "port" : "1099",
    "host" : "localhost",
    "queries" : [ {
      "obj" : "java.lang:type=Memory",
      "attr" : [ "HeapMemoryUsage", "NonHeapMemoryUsage" ],
      "resultAlias":"jvmMemory",
      "outputWriters" : [ {
        "@class" : "com.googlecode.jmxtrans.model.output.InfluxDbWriterFactory",
        "url" : "http://localhost:8086/",
        "database" : "mydb",
        "username" : "influxuser"
      } ]
    } ]
  } ]
}

saad373:
I fixed this by providing "retentionPolicy":"autogen" in influxDB configuration.

lewg:
[...]

PR #510, opened 2016-11-06
Allow InfluxDBWriter to connect to 1.0 out of the box

maxj:
This uses the latest version of influxdb-java (2.4) to support connecting to InfluxDB 1.0. 

This also changes the default retention policy from "default" to "autogen"

This is a breaking change for anyone relying on the old defaults.

Resolves errors:

```sh
ava.lang.RuntimeException: {"error":"error parsing query: found NOT, expected ; at line 1, char 20"}
```

and 

```sh
java.lang.RuntimeException: {"error":"retention policy not found: default"}
```

As mentioned in:
https://github.com/jmxtrans/jmxtrans/issues/489

### DIFF
--- a/jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterFactory.java
+++ b/jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterFactory.java
@@ -50,11 +50,11 @@ public class InfluxDbWriterFactory implements OutputWriterFactory {
 
 	/**
 	 * The deault <a href=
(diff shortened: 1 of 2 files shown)


gehel:
This looks good. I know mostly nothing about InfluxDB. Why change the default retention policy? I prefer to not introduce breaking changes unless there is a good reason to do it. Could you explain the choice a bit more?

Thanks!

maxj:
I am new to InfluxDB too, currently migrating from Graphite to Influx.

From the ticket referenced below, it looks like InfluxDB switched their default retention policy name from ```default``` to ```autogen``` in 1.0.  

https://github.com/influxdata/influxdb/issues/6825

Alternatively, it might be possible to set the retention policy in jmxtrans yml.  If that is the case, it would help newcomers use the project without breaking anything, I can take a look at that later this week.

gehel:
It seems to make sense to follow InfluxDB...

PR #515, opened 2016-11-10
null check and defaults added

peterwoj:
Added a default for mergeTypeNamesTags to match the docs as well as added a null check against typeNames since that item is supposed to be optional.

Made more changes to move around Server through OpenTSDBMessageFormatter and removed a couple getHostName() calls so that only Server resolves the hostname.  The property on outputWriters in the configuration now stays as a boolean value throughout.  Little larger change than the initial but tests for OpenTSDBWriter* and TCollectorUDPWriter* pass with the new configuration.

### DIFF
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/OpenTSDBWriter2.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/OpenTSDBWriter2.java
@@ -49,7 +49,7 @@ public class OpenTSDBWriter2 implements WriterBasedOutputWriter {
 	public void write(@Nonnull final Writer writer, @Nonnull Server server, @Nonnull Query query, @Nonnull Iterable<Result> results) throws IOException {
 
 		for(String resultString : messageFormatter.formatResults(results, server)) {
-            logger.trace(resultString);
+			logger.trace(resultString);
 			writer.write("put " + resultString + "\n");
 		}
 	}
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/OpenTSDBWriter2.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/OpenTSDBWriter2.java
@@ -46,7 +46,11 @@ public class OpenTSDBWriter2 implements WriterBasedOutputWriter {
 	}
 
 	@Override
-	public void write(@Nonnull final Writer writer, @Nonnull Server server, @Nonnull Query query, @Nonnull Iterable<Result> results) throws IOException {
+	public void write(
+			@Nonnull Writer writer,
+			@Nonnull Server server,
+			@Nonnull Query query,
+			@Nonnull Iterable<Result> results) throws IOException {
 
 		for(String resultString : messageFormatter.formatResults(results, server)) {
 			logger.trace(resultString);
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/OpenTSDBWriter2.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/OpenTSDBWriter2.java
@@ -27,6 +27,8 @@ import com.googlecode.jmxtrans.model.Result;
 import com.googlecode.jmxtrans.model.Server;
 import com.googlecode.jmxtrans.model.output.support.WriterBasedOutputWriter;
 import com.googlecode.jmxtrans.model.output.support.opentsdb.OpenTSDBMessageFormatter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.concurrent.ThreadSafe;
@@ -35,6 +37,7 @@ import java.io.Writer;
 
 @ThreadSafe
 public class OpenTSDBWriter2 implements WriterBasedOutputWriter {
+	private static final Logger logger = LoggerFactory.getLogger(OpenTSDBWriter2.class);
 
 	private final OpenTSDBMessageFormatter messageFormatter;
 
@@ -45,7 +48,8 @@ public class OpenTSDBWriter2 implements WriterBasedOutputWriter {
 	@Override
 	public void write(@Nonnull final Writer writer, @Nonnull Server server, @Nonnull Query query, @Nonnull Iterable<Result> results) throws IOException {
 
-		for(String resultString : messageFormatter.formatResults(results)) {
+		for(String resultString : messageFormatter.formatResults(results, server)) {
+            logger.trace(resultString);
 			writer.write("put " + resultString + "\n");
 		}
 	}
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriter.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriter.java
@@ -68,7 +68,7 @@ public class TCollectorUDPWriter extends OpenTSDBGenericWriter {
 			@JsonProperty("tagName") String tagName,
 			@JsonProperty("mergeTypeNamesTags") Boolean mergeTypeNamesTags,
 			@JsonProperty("metricNamingExpression") String metricNamingExpression,
-			@JsonProperty("addHostnameTag") Boolean addHostnameTag,
+			@JsonProperty("addHostnameTag") boolean addHostnameTag,
 			@JsonProperty("settings") Map<String, Object> settings) throws LifecycleException, UnknownHostException {
 		super(typeNames, booleanAsNumber, debugEnabled, host, port, tags, tagName, mergeTypeNamesTags, metricNamingExpression,
 				addHostnameTag, settings);
@@ -93,7 +93,7 @@ public class TCollectorUDPWriter extends OpenTSDBGenericWriter {
 	@Override
 	public void internalWrite(Server server, Query query, ImmutableList<Result> results) throws Exception {
 		this.startOutput();
-		for (String resultString : messageFormatter.formatResults(results)){
+		for (String resultString : messageFormatter.formatResults(results, server)){
 			log.debug("TCollectorUDP Message: {}", resultString);
 			this.sendOutput(resultString);
 		}
@@ -150,7 +150,7 @@ public class TCollectorUDPWriter extends OpenTSDBGenericWriter {
 		private String tagName;
 		private Boolean mergeTypeNamesTags;
 		private String metricNamingExpression;
-		private Boolean addHostnameTag;
+		private boolean addHostnameTag;
 
 		private Builder() {}
 
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
@@ -200,7 +200,7 @@ public class OpenTSDBMessageFormatter {
 				addTag(resultString, addTagName, addTagValue);
 			}
 
-			if ((typeNames != null) && (!typeNames.isEmpty())) {
+			if (!typeNames.isEmpty()) {
 				this.addTypeNamesTags(resultString, result);
 			}
 
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
@@ -46,7 +46,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.collect.FluentIterable.from;
 import static com.googlecode.jmxtrans.util.NumberUtils.isNumeric;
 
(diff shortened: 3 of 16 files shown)


jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/OpenTSDBWriter2.java
```diff
@@ -45,7 +48,8 @@ public OpenTSDBWriter2(OpenTSDBMessageFormatter messageFormatter) {
 	@Override
 	public void write(@Nonnull final Writer writer, @Nonnull Server server, @Nonnull Query query, @Nonnull Iterable<Result> results) throws IOException {
 
-		for(String resultString : messageFormatter.formatResults(results)) {
+		for(String resultString : messageFormatter.formatResults(results, server)) {
+            logger.trace(resultString);
```
gehel: Spaces should be tabs (I know, crazy rule, but let's keep coherent style)

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
```diff
@@ -64,6 +66,7 @@
 
 	private final boolean mergeTypeNamesTags;
 	private final String hostnameTag;
+	private Server server;
```
gehel: Server should not be part of the state of the message formater. The same message formater can be used to format messages from multiple different servers, and can do that concurrently.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
```diff
@@ -163,6 +172,10 @@ private void formatResultString(StringBuilder resultString, String metricName, l
 	}
 
 	public Iterable<String> formatResults(Iterable<Result> results) {
+		return formatResults(results, null);
+	}
+	public Iterable<String> formatResults(Iterable<Result> results, Server server) {
```
gehel: Adding `Server` to the arguments of `formatResults` seems to make sense. But it should replace the previous `formatArguments`, not override it (and you should also modify all call sites). Passing `server` indirectly, via a class property is not thread safe and mixes what class state with call specific parameters.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
```diff
@@ -193,7 +206,7 @@ protected void processOneMetric(List<String> resultStrings, Result result, Objec
 				addTag(resultString, addTagName, addTagValue);
 			}
 
-			if (!typeNames.isEmpty()) {
+			if ((typeNames != null) && (!typeNames.isEmpty())) {
```
gehel: `typeNames` is tagged `@NonNull` in the constructor arguments. We could be more explicit by tagging it `@NonNull` also in the class variables. But in the end, the result is the same, `typeNames` should never be null, so the null check is not needed. If there is a code path where this is not the case, that code path should be fixed.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
```diff
@@ -101,7 +104,13 @@ public OpenTSDBMessageFormatter(@Nonnull ImmutableList<String> typeNames,
 	 */
 	void addTags(StringBuilder resultString) {
 		if (hostnameTag != null) {
-			addTag(resultString, "host", hostnameTag);
+			if (server != null) {
+				addTag(resultString, "host",
+						firstNonNull(server.getAlias(), server.getHost()));
```
gehel: `firstNonNull(server.getAlias(), server.getHost()))` should be moved to the `Server` class. I'm not sure of the naming, but probably something like `getName()` or `getLabel()`. This same code is probably in quite a few other places.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriter.java
```diff
@@ -68,7 +68,7 @@ public TCollectorUDPWriter(
 			@JsonProperty("tagName") String tagName,
 			@JsonProperty("mergeTypeNamesTags") Boolean mergeTypeNamesTags,
 			@JsonProperty("metricNamingExpression") String metricNamingExpression,
-			@JsonProperty("addHostnameTag") Boolean addHostnameTag,
+			@JsonProperty("addHostnameTag") boolean addHostnameTag,
```
gehel: All that class hierarchy uses `Boolean addHostnameTag` (with a capital `B`). This makes clear that this parameter is optional. Changing it to `boolean` is probably not what we want.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriter.java
```diff
@@ -150,7 +150,7 @@ public static Builder builder() {
 		private String tagName;
 		private Boolean mergeTypeNamesTags;
 		private String metricNamingExpression;
-		private Boolean addHostnameTag;
+		private boolean addHostnameTag;
```
gehel: Should stay as `Boolean`.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/OpenTSDBGenericWriterTests.java
```diff
@@ -82,7 +85,7 @@ public void setupTest() {
 		Mockito.when(this.mockResult.getClassName()).thenReturn("X-DOMAIN.PKG.CLASS-X");
 		Mockito.when(this.mockResult.getTypeName()).
 				thenReturn("Type=x-type-x,Group=x-group-x,Other=x-other-x,Name=x-name-x");
-
+		Mockito.when(this.mockServer.getLabel()).thenReturn("myhostname");
```
gehel: It is probably not necessary to use a mock here. One of the fixture in `ServerFixtures` should work just as well.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriterTests.java
```diff
@@ -69,6 +71,7 @@ public void setupTest() throws Exception {
 		this.mockResult = Mockito.mock(Result.class);
 		this.mockDgSocket = Mockito.mock(DatagramSocket.class);
 		this.mockLog = Mockito.mock(Logger.class);
+		this.mockServer = Mockito.mock(Server.class);
```
gehel: No need for a mock.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatterTest.java
```diff
@@ -38,23 +40,27 @@
...
-	protected Result mockResult;
+	private Query mockQuery;
+	private Result mockResult;
+	private Server mockServer;
 
 	@Before
 	public void setupTest() {
 		this.mockQuery = Mockito.mock(Query.class);
 		this.mockResult = Mockito.mock(Result.class);
+		this.mockServer = Mockito.mock(Server.class);
```
gehel: No need for a mock.

gehel:
Good job! I still have a few comments, but this is really on a good way...

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriter.java
peterwoj: The default value here is to be false, so wouldn't that save a forced check later?  This also differs on the default value of true which is expected for openTSDB writes and documentation.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
```diff
@@ -63,19 +64,19 @@
...
+		this(typeNames, tags, DEFAULT_TAG_NAME, null, true, true);
 	}
 
 	public OpenTSDBMessageFormatter(@Nonnull ImmutableList<String> typeNames,
 									@Nonnull ImmutableMap<String, String> tags,
 									@Nonnull String tagName,
 									@Nullable String metricNamingExpression,
 									boolean mergeTypeNamesTags,
-									@Nullable String hostnameTag) throws UnknownHostException, LifecycleException {
+									Boolean hostnameTag) throws LifecycleException {
```
gehel: `OpenTSDBMessageFormatter` is an internal component of jmxtrans, where we can ensure that it is never called with `hostnameTag == null`. So I would prefer to have a `boolean` primitive here and not a `Boolean` object.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/OpenTSDBGenericWriterTests.java
```diff
@@ -67,6 +70,7 @@
 	public void setupTest() {
 		this.mockQuery = Mockito.mock(Query.class);
 		this.mockResult = Mockito.mock(Result.class);
+		this.server = ServerFixtures.dummyServer();
```
gehel: Using `ServerFixutures` also means that you don't need to keep server as instance variable, you can just inline it everywhere it is used.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriter2Test.java
```diff
@@ -39,13 +41,15 @@
...
 	private Result result;
+	private Server server;
 
 	@Before
 	public void setup() {
 		openTSDBMessageFormatter = Mockito.mock(OpenTSDBMessageFormatter.class);
 		writer = new TCollectorUDPWriter2(openTSDBMessageFormatter);
 		outputWriter = Mockito.mock(Writer.class);
 		result = Mockito.mock(Result.class);
+		server = ServerFixtures.dummyServer();
```
gehel: This instance variable can be inlined.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriterTests.java
```diff
@@ -69,6 +72,7 @@ public void setupTest() throws Exception {
 		this.mockResult = Mockito.mock(Result.class);
 		this.mockDgSocket = Mockito.mock(DatagramSocket.class);
 		this.mockLog = Mockito.mock(Logger.class);
+		this.server = ServerFixtures.dummyServer();
```
gehel: Instance variable can be inlined

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatterTest.java
```diff
@@ -38,24 +41,26 @@
...
-	protected Result mockResult;
+	private Query mockQuery;
+	private Result mockResult;
+	private Server server;
 
 	@Before
 	public void setupTest() {
 		this.mockQuery = Mockito.mock(Query.class);
 		this.mockResult = Mockito.mock(Result.class);
+		this.server = ServerFixtures.dummyServer();
```
gehel: Instance variable can be inlined.

gehel:
This looks fairly good! I'm still adding a few minor comments. The only one that I would really like to see fixed before merging is the boolean vs Boolean issues.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/support/opentsdb/OpenTSDBMessageFormatter.java
peterwoj: Makes sense no configuration here it's just brought down from the Writer.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/OpenTSDBGenericWriterTests.java
```diff
@@ -104,7 +102,7 @@ public void testEmptyTagSetting() throws Exception {
 		Assertions.assertThat(writer.getTypeNames()).isEmpty();
 
 		writer.start();
-		writer.doWrite(server, this.mockQuery, ImmutableList.of(this.mockResult));
+		writer.doWrite(ServerFixtures.dummyServer(), this.mockQuery, ImmutableList.of(this.mockResult));
```
gehel: `static import` might be more readable (really minor comment)

gehel:
Build is failing on some unused imports, but outside from that, this looks ready to be merged.

jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/TCollectorUDPWriter.java
peterwoj: I know, leaving it to coercion doesn't feel as clean.  Good news being is that the class is on the deprecation list 👍 , now to work on gaining adoption of the Factory.

peterwoj:
Odd I ran it locally before pushing and didn't see the problem, cleared out and sending again.

gehel:
Those checks are run on the validation phase. If you just ran `mvn package` instead of `mvn verify` they would not have been run. If you want to test even more, you can add the `withMutationTests` profile (`mvn -PwithMutationTests verify`), but it takes quite some time (and does generate nice report btw).

gehel:
Thanks a lot for all the work! And for being patient with my code related OCD...

peterwoj:
Not a problem, rather have a clean application that let general so so code in.  Thanks for accepting and having it getting it in the main line.

PR #516, opened 2016-11-11
new packaging

gehel:
I'm trying to put some order into .deb and .rpm packaging. This will
require a few iterations, probably some changes to jmxtrans itself
and reviews from people who actually understand all that packaging
stuff...

Definitely work in progress!


gehel:
@elukey if you want to have a look, I started rearranging the modules around packaging. I have a few idea, but not sure what make sense and what doesn't. In no particular order:

* have a `-common` package with the jar and minimal script to start jmxtrans manually
* have different packages for systemd systems vs upstart / sysv
* ...

PR #531, opened 2016-12-14
Handle MBean TabularData in statsd writer

itzg:
Using the following query results in a BufferOverflowException due to an attempt to write [a very long line](https://gist.github.com/itzg/7bd028867c5df5cbc6839386d9278baa#file-example-of-very-long-tabulardata-statsdwriter2-line-txt) to the statsd channel.

```
{
  "obj": "java.lang:type=GarbageCollector,name=G1 Young Generation",
  "resultAlias": "java.lang.type-GarbageCollector.name-G1_Young_Generation",
  "attr": [
    "LastGcInfo"
  ],
  "outputWriters": [
    {
      "@class": "com.googlecode.jmxtrans.model.output.StatsDWriterFactory",
      "bucketType": "g",
      "host": "localhost",
      "port": 8125,
      "rootPrefix": "cassandra.jmx",
      "stringValuesAsKey": true
    }
  ]
}
```

I narrowed the root cause of that exception to the line rendered [here](https://github.com/jmxtrans/jmxtrans/blob/757e06a32fa218a7711814f4fd92fca90f4b10ce/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/StatsDWriter2.java#L87) due to the inclusion of the `TabularDataSupport` object in the results values [here](https://github.com/jmxtrans/jmxtrans/blob/757e06a32fa218a7711814f4fd92fca90f4b10ce/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java#L139).

Since `stringValuesAsKey` is set to `true`, [this logic](https://github.com/jmxtrans/jmxtrans/blob/757e06a32fa218a7711814f4fd92fca90f4b10ce/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/StatsDWriter2.java#L93) allows the TabularData object to be considered "numeric" or a "string value".

This PR proposes a solution which is to further tighten down the `isNotValidValue` logic in `StatsDWriter2`. A new unit test has been introduced to approximate the results of the query declaration above where primitive values need to pass, but the tabular data excluded.

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
@@ -136,7 +136,7 @@ public class JmxResultProcessor {
 			if (value instanceof TabularDataSupport) {
 				TabularDataSupport tds = (TabularDataSupport) value;
 				processTabularDataSupport(accumulator, attributeName + "." + key, tds);
-				values.put(key, value);
+				// continue because we added tabular contents within above, but need primitives at this level
 			} else if (value instanceof CompositeDataSupport) {
 				// now recursively go through everything.
 				CompositeDataSupport cds2 = (CompositeDataSupport) value;
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/ResultFixtures.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/ResultFixtures.java
@@ -25,6 +25,12 @@ package com.googlecode.jmxtrans.model;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
+import javax.management.openmbean.*;
+import java.lang.management.MemoryUsage;
+import java.util.Collections;
+import java.util.Map;
+import java.util.TreeMap;
+
 public final class ResultFixtures {
 	private ResultFixtures() {}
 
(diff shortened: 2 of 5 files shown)


gehel:
Thanks a lot for the investigation and the fix!

Tabular data should be processed into individual results. [This part of the code](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java#L153-L181) is messy and not tested enough, but is probably close to where the root of the issue lies.

itzg:
Yes, the call to `processTabularDataSupport` seems to be correctly taking care of the individual metric extraction (messy as it is :) ). I also pondered removing [this line](https://github.com/jmxtrans/jmxtrans/blob/757e06a32fa218a7711814f4fd92fca90f4b10ce/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java#L139), but worried that might have ramifications on other output writers. Instead I thought excluding within the statsd writer would minimize ripple effects.

BTW, I'm guessing we all haven't noticed this before because the G1 GC seems to have introduced a ton of composite metrics just within those one or two attributes.

itzg:
I'm working on the checkstyle-build failures...

gehel:
It looks to me that removing [JmxResultProcessor:139](https://github.com/jmxtrans/jmxtrans/blob/757e06a32fa218a7711814f4fd92fca90f4b10ce/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java#L139) is a better / cleaner solution. I'm pretty sure that we don't have any OutputWriter trying to do smart things with tabular data... nor should they.

As a side note, [Result.values](https://github.com/jmxtrans/jmxtrans/blob/757e06a32fa218a7711814f4fd92fca90f4b10ce/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java#L54) should have a much stricter type than `Object`. The OutputWriters themselves should not have to deal with the complexity of JMX...

itzg:
Yeah, that makes more sense since it would better mirror [the code below it](https://github.com/jmxtrans/jmxtrans/blob/757e06a32fa218a7711814f4fd92fca90f4b10ce/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java#L140-L144).

Hmm, I'll probably re-work the unit test to wrap `JmxResultProcessor` then, since that'll be the altered logic under test...

gehel:
Yeah, that `values.put(key, value);` is most probably a left over from a bad refactoring a long time ago. And it was not catched before because mostly no one uses tabular data...

itzg:
...nice, I'll just add more assertions in [this test](https://github.com/jmxtrans/jmxtrans/blob/0b8bd70102791c8047c0d44f05a0e8639a944f25/jmxtrans-core/src/test/java/com/googlecode/jmxtrans/jmx/JmxResultProcessorTest.java#L154-L154)

itzg:
@gehel this is now ready for a re-review

...and [this build](https://travis-ci.org/jmxtrans/jmxtrans/builds/184294910) seems to be stuck

gehel:
Thanks a lot!

itzg:
Thanks for merging my PRs. Can I talk you into cutting a 261 release soon? This PR resolves the last of the issues we saw in our deployment.

gehel:
Release in progress: https://travis-ci.org/jmxtrans/jmxtrans/builds/184360956

PR #538, opened 2016-12-21
FileWriter Atomic Rename Improvement

luismfonseca:
After running this for a while, I realised that `StandardCopyOption.REPLACE_EXISTING` will not use atomic moves even if they are possible. The `StandardCopyOption.ATOMIC_MOVE` will throw an exception if the underlying OS doesn't support atomic moves.

The `renameTo` function will do an atomic move if possible, and fall back to non-atomic if not possible.


gehel:
Some unused imports need to be cleaned up.

luismfonseca:
I think this is good to go now. @gehel

gehel:
@luismfonseca Thanks!

Issue #539, opened 2016-12-22
Filter out Non-Number Numbers


kevinconaway:
There are some JMX metrics that will return numbers like `Double.NaN`, `Double.POSITIVE_INFINITY` or `Double.NEGATIVE_INFINITY`.  These metrics will pass the `NumberUtils.isNumeric` check because it simply does an `instanceof Number` check.  However these values can't really be processed by Graphite (or perhaps any other down stream system, but I can't speak for those) so I think they should be filtered out.

@gehel thoughts here?  Happy to submit a PR if needed

gehel:
@kevinconaway: yes, a PR would be great! I don't know if we currently have any backend that would support those "almost number" values, but it seems likely that some backend out there does support them (even if jmxtrans does not have integration with them yet). So fixing this in the GraphiteWriter seems to make sense. Extracting a utility method that could be reused in other writers would be great!

kevinconaway:
Thanks for merging this so quickly.  It would be good if a release could be made with these changes when you have time.

Thanks!

gehel:
Release in progress: https://travis-ci.org/jmxtrans/jmxtrans/builds/186118029

itzg:
FWIW, this is similar to the issue I faced+fixed with the statsd writer https://github.com/jmxtrans/jmxtrans/pull/535 , but in that case their protocol had a way to convey a "non number".

tanuj83:
do we have solution for Kafka as well ?  #751

PR #540, opened 2016-12-22
Issue #539 Skip invalid "numbers" when reporting to graphite

kevinconaway:
@gehel This PR fixes the issue reported in #539 

I added a utility method `isValidNumber` that ensures that the number is not NaN or infinity.

### DIFF
--- a/jmxtrans-utils/src/main/java/com/googlecode/jmxtrans/util/NumberUtils.java
+++ b/jmxtrans-utils/src/main/java/com/googlecode/jmxtrans/util/NumberUtils.java
@@ -43,4 +43,26 @@ public final class NumberUtils {
 		return false;
 	}
 
+	public static boolean isValidNumber(Object value) {
+		if (!(value instanceof Number)) {
+			return false;
+		}
+
+		Number number = (Number) value;
+
+		if (number instanceof Double) {
+			if (Double.isNaN(number.doubleValue()) || Double.isInfinite(number.doubleValue())) {
+				return false;
+			}
+		}
+
+		if (number instanceof Float) {
+			if (Float.isNaN(number.floatValue()) || Float.isInfinite(number.floatValue())) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
 }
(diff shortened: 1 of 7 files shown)


jmxtrans-utils/src/main/java/com/googlecode/jmxtrans/util/NumberUtils.java
```diff
@@ -43,4 +43,26 @@ public static boolean isNumeric(Object value) {
 		return false;
 	}
 
+	public static boolean isValidNumber(Object value) {
```
gehel: `value` is nullable, so a null check is probably a good thing. If I remember correctly, `(((Object) null) instanceof Number) == false) but that's non obvious...

jmxtrans-utils/src/main/java/com/googlecode/jmxtrans/util/NumberUtils.java
```diff
@@ -43,4 +43,26 @@ public static boolean isNumeric(Object value) {
...
+		if (!(value instanceof Number)) {
```
gehel: A number represented as a `String` is still treated as a number for most OutputWriters (including Graphite). We are probably missing a test case about this.

jmxtrans-utils/src/main/java/com/googlecode/jmxtrans/util/NumberUtils.java
kevinconaway: Correct.  instanceof checks for null always return false.  I can make it explicit if you wish.

jmxtrans-utils/src/main/java/com/googlecode/jmxtrans/util/NumberUtils.java
kevinconaway: True.  I had really focused this on instances of Numbers that would be returned from JMX, not a string that would be converted.

I can address this.

kevinconaway: Although the string representations of NaN and infinity don't currently pass the existing `IsNumeric`

gehel:
Thanks a lot!

PR #543, opened 2016-12-27
kafka output writer does not close connection after producing to a topic

frederiko:
This PR fixes a bug found when shipping a significant amount of metrics. Upon producing a message to a topic, it appears that the Kafka producer does not close the connection. In cases where a significant amount of messages are sent, connection are kept in CLOSED_WAIT state for a long.

### DIFF
--- a/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
+++ b/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.concurrent.NotThreadSafe;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
```diff
@@ -117,14 +119,17 @@ protected void internalWrite(Server server, Query query, ImmutableList<Result> r
...
 					}
 				} else {
 					log.warn("Unable to submit non-numeric value to Kafka: [{}] from result [{}]", value, result);
 				}
 			}
 		}
+
+		// Batch them all
+		producer.send(messages);
+		producer.close();
```
gehel: Looking at the code in `kafka.javaapi.producer.Producer` it looks like if the producer is shutdown, it does not restart. Since we reuse the same producer for the lifetime of our `KafkaWriter`, this does not seem like a good idea. The `Producer` should probably be closed only when the `KafkaWriter` is closed (by overriding the `close()` method). It also looks like the `Producer` class is deprecated and we should replace it with `KafkaProducer`.

That being said, I have no experience with Kafka, so I might be completely wrong...

Sorry for the delay in my review!

PR #551, opened 2017-03-29
Fix: ObjectAlreadyExistsException because one already exists

yyvess:
Triggers name are  generated with currentTimeMillis() and in some case two triggers can have same generated name.

That cause the following error on the service loading :

> ERROR (com.googlecode.jmxtrans.JmxTransformer:190) - Error scheduling job for server: 
> Caused by: org.quartz.ObjectAlreadyExistsException: Unable to store Trigger with name: 'test.nespresso.com:9200-1490623007618' and group: 'DEFAULT', because one already exists with this identification.

I have replace  currentTimeMillis() by nanoTime() to fix this issue.

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
@@ -387,37 +387,39 @@ public class JmxTransformer implements WatchedCallback {
 	}
 
 	private void scheduleJob(Server server) throws ParseException, SchedulerException {
-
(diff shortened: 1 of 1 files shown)


gehel:
Thanks for the fix! Longer term, we need to get rid of quartz, a simple ScheduledExecutorService would be much simpler...

gehel:
@yyvess let me know if you need a new release with that fix...

blaiseb:
Thanks for the merge @gehel ! I was the one who asked @yyvess for this fix. If you can do a new release we would greatly appreciate

gehel:
Minor issue with the build and PGP keys that needs to be fixed, but I'll release ASAP...

gehel:
Ok, PGP key was expired, this should now work better...

gehel:
release is done, should be available on maven central soon...

yyvess:
Thanks !!

PR #566, opened 2017-04-26
Add circuit breakers on Server and OutputWriters

### LINKED ISSUE #565: queryProcessorExecutor to have not more than two ProcessQueryThread tasks for one server+query
Hi!
I'm observing logs flooded with error messages like this:
Could not submit query {}. You could try to size the 'queryProcessorExecutor' to a larger size.
(JmxUtils#processServer)
queryProcessorExecutor has got a limit of 100000 tasks and it seems that they cannot be executed before next ServerJob.execute is triggered.

This also results in a very small chance for scheduling a full list of ProcessQueryThread tasks for each server when some server is responding slower than others. The executor queue is full of slow tasks and prevents others to be enqueued.

Isn't it better to have something that prevents having more than, say, two similar ProcessQueryThread tasks in queryProcessorExecutor?

I'm far from producing a good code but this fix seems to solve my problem:
[jmx.zip](https://github.com/jmxtrans/jmxtrans/files/954533/jmx.zip)

gehel:
This shoudl help in the case where a server or a backend becomes unresponsive.
The circuit breaker will open and no time will be wasted in trying to contact
a node that is down.

fixes #565.

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
@@ -24,9 +24,9 @@ package com.googlecode.jmxtrans.model;
 
 import com.fasterxml.jackson.annotation.JacksonInject;
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -40,6 +40,9 @@ import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
 import lombok.experimental.Accessors;
+import net.jodah.failsafe.CircuitBreaker;
+import net.jodah.failsafe.FailsafeException;
+import net.jodah.failsafe.function.CheckedRunnable;
 import org.apache.commons.pool.KeyedObjectPool;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


gquintana:
This is brilliant! 
What's missing to merge this PR?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -256,9 +264,43 @@ private Server(
 		this.pool = checkNotNull(pool);
 		this.outputWriterFactories = ImmutableList.copyOf(firstNonNull(outputWriterFactories, ImmutableList.<OutputWriterFactory>of()));
 		this.outputWriters = ImmutableList.copyOf(firstNonNull(outputWriters, ImmutableList.<OutputWriter>of()));
+		this.breaker = new CircuitBreaker()
+				.withFailureThreshold(3, 10)
```
gquintana: Timeouts and other timings should be configurable

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -256,9 +264,43 @@ private Server(
...
+				.withSuccessThreshold(1)
+				.withDelay(1, MINUTES)
+				.withTimeout(250, MILLISECONDS)
+				.onOpen(createLogger("Circuit breaker opened for server {}", this, false))
+				.onClose(createLogger("Circuit breaker closed for server {}", this, true))
+				.onHalfOpen(createLogger("Circuit breaker half opened for server {}", this, true));
+	}
+
+	private CheckedRunnable createLogger(final String message, final Server server, final boolean isSuccess) {
+		return new CheckedRunnable() {
```
gquintana: This anonymous class is the same as in CircuitBreakerOutputWriter.
Instead of true/false flag, using the log level would be clearer, yet SLF4J doesn't allow logging with variable log levels.

gehel:
@gquintana thanks for the kind words! Honestly, beside a rebase, I'm not sure what's missing... It might actually work as-is... (plus the comments you added). I mainly need to find time to get back to it. But feel free to take over!

gehel:
If I remember correctly, the real missing piece is exposing the configuration as command line flags. Not that hard, but that part is also messy. We should at some point move to a real configuration file, there are too many CLI options already...

PR #568, opened 2017-04-29
567 - fix influx output writer to properly use typeNames 

fksimon:
Currently when using wildcards with influxdb output writer it does not use the specified typeNames at all, which causes non-unique names and errors.   Update to use typeNames properly, they are prepended to the key name to make a unique key.


gehel:
Looks good! Thanks for the fix!

PR #573, opened 2017-06-15
change jmxtrans.conf default path

tiny1990:
in https://github.com/jmxtrans/jmxtrans/wiki/Installation#notice 
use /etc/sysconfig/jmxtrans.conf  better than /etc/sysconfig/jmxtrans  if CONF_FILE  not set


gehel:
It looks that at least in RedHat, the files in `/etc/sysconfig` do not end with `.conf`. So the current script seems to follow the standard practice. Is there an underlying issue that still needs to be fixed?

tiny1990:
thanks,I am not good at linux shell

gehel:
No problem! Thanks for the PR anyway!

PR #575, opened 2017-06-19
Change default log level to INFO

PierreF:
On Ubuntu/Debian, the default log level is DEBUG, which I think is too verbose. Most user don't need debug level, and if needed it's easy to change the default on their setup.
My log file is about 700Mb per day (I don't monitor lots of metrics - about 26 metrics, but I do it every 10 seconds).

This PR change the default on .deb package from DEBUG to INFO.
Note: .rpm package are already at info level


gehel:
Thanks a lot! This change definitely make sense!

PR #576, opened 2017-06-19
Add purge of old logs file

PierreF:
It seems that currently nothing purge old logs file. Which means that after a while a lots of files may be present in /var/log/jmxtrans and consume disk space.

I'm not sure logs older that few weeks are still useful.

The PR add a daily cron script that remove all archived logs older than 32 days.

### DIFF
--- a/jmxtrans/cron.daily/jmxtrans
+++ b/jmxtrans/cron.daily/jmxtrans
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+find /var/log/jmxtrans -type f -mtime +32 -name "jmxtrans-*.log.gz" -print0 | xargs -0 rm -f
--- a/jmxtrans/pom.xml
+++ b/jmxtrans/pom.xml
@@ -393,6 +393,14 @@
 									</sources>
 								</mapping>
 								<mapping>
+									<directory>${package.install.dir}/tools</directory>
+									<sources>
+										<source>
+											<location>${project.basedir}/tools</location>
+										</source>
+									</sources>
+								</mapping>
+								<mapping>
 									<directory>/usr/bin</directory>
 									<directoryIncluded>false</directoryIncluded>
 									<filemode>755</filemode>
@@ -405,12 +413,13 @@
 									</sources>
 								</mapping>
 								<mapping>
-									<directory>/etc/${package.daemon.name}</directory>
+									<directory>/etc/cron.daily</directory>
 									<configuration>true</configuration>
+									<filemode>755</filemode>
 									<sources>
-										<softlinkSource>
-											<location>${package.install.dir}/etc</location>
-										</softlinkSource>
+										<source>
+											<location>${project.basedir}/cron.daily</location>
+										</source>
 									</sources>
 								</mapping>
 								<mapping>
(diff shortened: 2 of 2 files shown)


gehel:
Thanks for the PR! For log rotation, I'd much prefer have this managed by the logging framework (log4j / logback, yes, there is a mess there). That's one less system dependency...

PierreF:
I'm not a Java developer, so I've a very little knowledge of log4j 1.x/log4j 2.x/logback and co.
But I initially tried to look at handling this on logging framework (currently log4j 1.x) and it seems that it does only support it with size based rotation.
With a time based rotation logback seems to support it. And (I'm not sure about this one) log4j 2.x.

So for me, if we want to support this in logging framework, we need to add logback and/or upgrade to log4j 2.x.

Should I try to look at upgrading log4j or add logback ?

gehel:
Since you seem to be managing Java application, some knowledge of the standard logging framework will probably help you in the long run :) (sorry I could not resist).

The situation is a bit complex... We have a mess of both logback and log4j. The way to go is definitely logback, but there are some OutputWriters that depend directly on log4j, which makes removing it (or upgrading it) a giant pain. Not that there is a [time based rolling policy](https://logging.apache.org/log4j/extras/apidocs/org/apache/log4j/rolling/TimeBasedRollingPolicy.html) in log4j-extra, which is already bundled with jmxtrans. This could be a short term approach.

PierreF:
From my search the logger that support deletion of history are:
* log4j 1.2 but with size based rotation: http://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/RollingFileAppender.html => maxBackupIndex
* log4j 2.8, using DefaultRolloverStrategy that has delete policy. Here we can keep a time based rotation and compressed archive
* logback that has TimeBasedRollingPolicy with rolling policy.

For the last solution, I don't see how to do it. jmxtrans-output-log4j use SLF4F and its log4j12 binding. From SLF4J documentation we should "drop one and *only one* binding [...] onto the [...] class path". Therefor I can't add logback binding without providing two binding.

The second solution means dropping usage of SLF4J (or we have the same problem as described above). It should be much more than just changing all logger creation. It also means that SLF4J + log4j 1.2 will still be used but jmxtrans-output-log4j. But since log4j 2 use another package namespace (org.apache.logging.log4j vs org.apache.log4j it should be conflict).

The first solution is just configuration change. But it means that rotation will be size based and no longer time based.

gehel:
Another solution would be to use https://logging.apache.org/log4j/extras/apidocs/org/apache/log4j/rolling/TimeBasedRollingPolicy.html with log4j. The class is already on the classpath.

PierreF:
I don't see how it help. It should already be used by [log4j.xml](https://github.com/jmxtrans/jmxtrans/blob/3dc555c34e2e5b48fc3adb541bb2ab6ef1088fc9/jmxtrans/src/main/resources/log4j.xml#L46) and don't seem to have any option to limit history size.
If you are thinking to https://logging.apache.org/log4j/extras/apidocs/org/apache/log4j/rolling/RollingFileAppender.html (which is also used by log4j.xml config) and its MaxBackupIndex option, it's a unused option. [Code](https://github.com/apache/log4j-extras/blob/trunk/src/main/java/org/apache/log4j/rolling/RollingFileAppender.java#L206) confirm that it's only a JavaDoc comment.

PierreF:
(closed my mistake, sry)

gehel:
In any case, the "right" solution would be to deprecate the log4j based output writers (keep them as an optional module) and move all logging to logback...

gquintana:
I also tried to replace the Log4J1 dependency by a Log4J2 dependency for similar need.
But the jmxtrans-output-log4j module has a strong dependency on Log4J1.

gehel:
Yeah, those log4j appenders are a mess. We should deprecate them and replace them by slf4j output writer, which can do the same things, but driven by logging configuration. This would require writing some docs on how to use it, remove the log4j appender from the final build (we could keep them around, just in case someone really likes them). My preference would be to migrate to logback more than log4j2, but I'm open to alternatives.

gquintana:
Since  v270 (maybe v269), JMXTrans uses Logback library. Its RollingFileAppender supports a [TimeBasedRollingPolicy](https://logback.qos.ch/manual/appenders.html#TimeBasedRollingPolicy) which supports both time and size rotation.

PR #579, opened 2017-06-27
Use new Kafka Producer API

gquintana:
The new Kafka Producer has fewer dependencies and is 100% Java

### DIFF
--- a/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
+++ b/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
@@ -35,9 +35,15 @@ import com.googlecode.jmxtrans.model.Server;
 import com.googlecode.jmxtrans.model.ValidationException;
 import com.googlecode.jmxtrans.model.output.BaseOutputWriter;
 import com.googlecode.jmxtrans.model.output.Settings;
-import kafka.javaapi.producer.Producer;
-import kafka.producer.KeyedMessage;
-import kafka.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+
+import static org.apache.kafka.clients.producer.ProducerConfig.BOOTSTRAP_SERVERS_CONFIG;
+import static org.apache.kafka.clients.producer.ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG;
+import static org.apache.kafka.clients.producer.ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG;
+
+import org.apache.kafka.common.serialization.StringSerializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
```diff
@@ -92,10 +98,10 @@ public KafkaWriter(
 						DEFAULT_ROOT_PREFIX);
 		// Setting all the required Kafka Properties
 		Properties kafkaProperties =  new Properties();
-		kafkaProperties.setProperty("metadata.broker.list", Settings.getStringSetting(settings, "metadata.broker.list", null));
-		kafkaProperties.setProperty("zk.connect", Settings.getStringSetting(settings, "zk.connect", null));
-		kafkaProperties.setProperty("serializer.class", Settings.getStringSetting(settings, "serializer.class", null));
-		this.producer= new Producer<>(new ProducerConfig(kafkaProperties));
+		kafkaProperties.setProperty(BOOTSTRAP_SERVERS_CONFIG, Settings.getStringSetting(settings, BOOTSTRAP_SERVERS_CONFIG, null));
```
gehel: It looks like we are breaking the compatibility with the previous implementation (by changing the name of some of the settings). If that is actually the case (I have no checked the value of those constants), it is probably the right time to completely remove this ugly settings map and move to explicit parameters on the constructor.

gehel:
Thanks! This looks like an interesting change! I'd like to take some more time to go over it, but I'll try to do that tomorrow. Thanks for your patience!

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
gquintana: Whatever the default settings may be, moving from the Scala producer to the Java producer breaks compatibility as settings are completely different.

About the default serializer I added, it makes sense because the record is of type <String,String>.

About moving from settings map to explicit parameters, IMHO it's not a good idea. At the moment, this class only exposes the mandatory settings, but there are many more which are useful (acks, retries, bulk size, SSL, SASL auth...). In the end, it would mean adding many fields.

gehel: If we want to keep configuration open ended, then we should at least have something explicit for required parameters and maybe expose a Properties object for the additional properties.

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
```diff
@@ -92,10 +98,10 @@ public KafkaWriter(
...
+		kafkaProperties.setProperty(KEY_SERIALIZER_CLASS_CONFIG, Settings.getStringSetting(settings, KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName()));
+		kafkaProperties.setProperty(VALUE_SERIALIZER_CLASS_CONFIG, Settings.getStringSetting(settings, VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName()));
```
gehel: Does it make sense to have the serializer configureable? It looks to me like the only reasonable value is StringSerializer (but I don't know much about Kafka).

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
gehel: What will happen if `BOOTSTRAP_SERVERS_CONFIG` is `null`? It looks like this should be a required parameter with no default value.

gquintana: How do you tell it's required?

gquintana: I agree, this solution would be ideal. 

How can I bind field of type Properties with something coming from the JSON/YAML configuration? Is there an example somewhere else in the code base?

gquintana:
Another problem I have with this Kafka writer, is that a new instance is created for each query. As each instance creates a new Kafka producer which in turn creates a thread (messages are sent in background). In the end I have too much threads. 

I wonder how I could reuse the same Kafka producer for all queries (provided writer settings are identical).

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter.java
gehel: There is no example yet. My guess is that [jackson-databind](https://github.com/FasterXML/jackson-databind) will make it "just work", but that requires testing.

gehel: Not sure I understand the question. I can tell it is required, because without a kafka server configured, the messages won't go anywhere. To make the code explicit about this requirement, a validation check should be added, which would raise an exception if the value isn't set.

gehel:
The OutputWriters are deduplicated, based on their equals / hashCode implementation. [Lombok](https://projectlombok.org/features/EqualsAndHashCode) can help you with the implementation.

Note that this Kafka writer still extends BaseOutputWriter. Splitting the OutputWriter and OutputWriterFactory implementation would be cleaner / nicer.

gquintana:
Closing as this PR is superseded by #586

gehel:
Thanks for the cleanup!

PR #580, opened 2017-06-28
Allow SSL connections to JMX server

gquintana:


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -306,6 +317,17 @@ private Server(
 			environment.put(JMXConnector.CREDENTIALS, credentials);
 		}
 
+		if (ssl) {
```
gehel: An integration test for this would be really welcomed! But it is probably non-trivial to write...

gquintana: I though about it... but it's not easy

gehel: If you find a way, you can always add that integration test later :)

gquintana:
Documentation added https://github.com/jmxtrans/jmxtrans/wiki/JMXRMI-over-SSL

PR #582, opened 2017-06-29
Fix bug in SSL and add IT test case

gquintana:
**JMX over SSL is broken on the master branch at the moment, it's my fault, the bug is the missing `this.ssl = ssl` in `Server`.**

I added an integration test case based on `JmxTransformerIT`.
I have to start some kind of JmxTrans server in external process so as to be able to tune JVM SSL options.

The integration test case is successful when you at the logs, but it fails because the assertion can not read the stdout of the JmxTrans process. Any advice is welcome...


gquintana:
I fixed the test case by redirecting output to file and checking the log file content

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/JmxTransDummyApp.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private File file(String filename) throws URISyntaxException {
+        return new File(getClass().getClassLoader().getResource(filename).toURI());
+    }
+
+    public void stop() throws LifecycleException {
+        jmxTransformer.stop();
+    }
+
+    public static void main(String[] args) throws LifecycleException, URISyntaxException, InterruptedException {
```
gehel: That main method is only called in SslJmxTransformerIT, but the fork used there should be replaced by running jmxtrans in process (see other comment).

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/JmxTransDummyApp.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.inject.Injector;
+import com.googlecode.jmxtrans.cli.JmxTransConfiguration;
+import com.googlecode.jmxtrans.exceptions.LifecycleException;
+import com.googlecode.jmxtrans.guice.JmxTransModule;
+
+import java.io.File;
+import java.net.URISyntaxException;
+
+public class JmxTransDummyApp {
```
gehel: Since this is the actual code under test for JmxTransformerIT, I'm not sure that moving it as a top level class make sense. I understand the intent of extracting that class, but maybe extract it as an inner class of JmxTransformerIT?

gehel: I'm not a big fan of the naming. the `App` suffix let me thing that this is a standalone application, run as its own process, when it is in reality just a Java component, running in the same JVM as the integration test.

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/SslJmxTransformerIT.java
```diff
@@ -0,0 +1,70 @@
...
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+@Category({IntegrationTest.class, RequiresIO.class})
+public class SslJmxTransformerIT {
+    private final File jmxTransAppStdOut = new File("target" + File.separator + "ssl-jmxtrans-out.log");
+    @Rule
+    public ExternalApp app = new ExternalApp(DummyApp.class).enableJmx(12346)
+            .enableSsl("localhost.jks", "localhost");
+    @Rule
+    public ExternalApp jmxTransApp = new ExternalApp(JmxTransDummyApp.class, "ssl-integration-test.json", "5")
```
gehel: You probably don't want to run the code under test (JmxTransDummyApp) as an external app. You still need to configure a JKS for the test, but you can do that in process.

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/SslJmxTransformerIT.java
```diff
@@ -0,0 +1,70 @@
...
+    private class JmxTransStdoutContains implements Callable<Boolean> {
+        private final String value;
+
+        public JmxTransStdoutContains(String value) {
+            this.value = value;
+        }
+
+        @Override
+        public Boolean call() throws Exception {
+            return Files.toString(jmxTransAppStdOut, Charset.forName("UTF-8")).contains(value);
```
gehel: Guava's `Charsets.UTF_8` should probably be used in place of the literal "UTF-8" (note that this code will disappear if we run jmxtrans in process).

jmxtrans-test-utils/src/main/java/com/googlecode/jmxtrans/test/ExternalApp.java
```diff
@@ -0,0 +1,146 @@
...
+import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+import static com.google.common.collect.FluentIterable.from;
+import static java.util.Arrays.asList;
+
+public class ExternalApp extends ExternalResource {
```
gehel: I think I'd prefer to keep this as a specific MonitorableApp instead of the now more generic ExternalApp (unless we really have a need for the generic one, which does not seem to be the case at this point).

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/JmxTransDummyApp.java
gquintana: I don't understand your comment, can you detail it bit more?

jmxtrans-test-utils/src/main/java/com/googlecode/jmxtrans/test/ExternalApp.java
gquintana: You still have a MonitorableApp, it's DummyApp! The purpose of ExternalApp (and the reason why I renamed the class), it that the aim of this class is now start an external Java process, may it be a monitored app or a monitoring app.
Do you suggest a class MonitorableApp calling ExternalApp starting the DummyApp ?

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/JmxTransDummyApp.java
gehel: Re-reading myself, I agree, I was not very clear :) Let's try again...

You should just remove this `main()` method and not run jmxtrans in another process.

gquintana: IMHO I can't do that because I need to configure the truststore of JmxTrans (with `.enableSsl("localhost.jks", "localhost")` in the IT) because the private key stored in the localhost.jks is self-signed.

Can you see a better solution?

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/SslJmxTransformerIT.java
gquintana: Files.toString method takes a Charset as argument not a String
https://google.github.io/guava/releases/19.0/api/docs/src-html/com/google/common/io/Files.html#line.273

I am going to replace Charset.forName by Charsets.UTF-8 it it what you want?

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/JmxTransDummyApp.java
gquintana: *It is* used as standalone app in SslJmxTransIT because I need to configure the SSL truststore at the JVM level. This app is started here
https://github.com/jmxtrans/jmxtrans/pull/582/files#diff-26c08ae67e6f2f71134e11a2082e7ae4R48

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/SslJmxTransformerIT.java
gquintana: *you can do that in process* How?

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/JmxTransDummyApp.java
gehel: Yep, my point being that we should not run it as an external app.

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/JmxTransDummyApp.java
gehel: Yes, you should be able to configure the client trust store in process (`System.setProperty("javax.net.ssl.trustStore", "localhost.jks");` - I have not tested that it actually works, so I'm probably wrong, but worth a try...).

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/SslJmxTransformerIT.java
gehel: Ok, that would have been my first try. I'm pretty sure I did something similar once a long time ago, but my memory is not all that good :(  I'll dig a bit and see if I can find a solution.

gehel: Probably something to do with [SSLContext](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLContext.html#setDefault(javax.net.ssl.SSLContext)). It might get messy and your approach of forking a new JVM might make more sense...

jmxtrans-test-utils/src/main/java/com/googlecode/jmxtrans/test/ExternalApp.java
gehel: No, I was suggesting that we keep only a single app (MonitorableApp).

gehel:
Damn, I let this one sleep for too long, We are now in merge conflict. The code itself looks good enough to me. I'll try to find some time to do the rebase... but not right now.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -512,6 +512,7 @@ private Builder(Server server) {
 			this.runPeriodSeconds = server.runPeriodSeconds;
 			this.numQueryThreads = server.numQueryThreads;
 			this.local = server.local;
+			this.ssl = server.ssl;
```
gquintana: Without this code on master is broken

gehel:
Thanks for the merge! For next time, I'd actually would have preferred, that way, we can keep the history of master as simple as possible. I'll just let Travis do its job and I'll merge this.

gquintana:
Documentation added https://github.com/jmxtrans/jmxtrans/wiki/JMXRMI-over-SSL

gquintana:
@gehel  I know it is the holidays season, but is there any news on this?

PR #586, opened 2017-06-30
New Kafka Writer

gquintana:
I tried to split Writer from WriterFactory: is this what you're expecting?
Can you give me some feedback about the design?

### DIFF
--- a/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/DefaultResultSerializer.java
+++ b/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/DefaultResultSerializer.java
@@ -50,64 +50,64 @@ import static com.googlecode.jmxtrans.util.NumberUtils.isNumeric;
  * Original Result serializer (looks like OpenTSDB data format)
  */
 public class DefaultResultSerializer implements ResultSerializer {
-    private static final Logger log = LoggerFactory.getLogger(DefaultResultSerializer.class);
-    private final JsonFactory jsonFactory;
-    private final ImmutableList<String> typeNames;
-    private final boolean booleanAsNumber;
-    private final String rootPrefix;
-    private final ImmutableMap<String, String> tags;
+	private static final Logger log = LoggerFactory.getLogger(DefaultResultSerializer.class);
+	private final JsonFactory jsonFactory;
+	private final ImmutableList<String> typeNames;
+	private final boolean booleanAsNumber;
+	private final String rootPrefix;
+	private final ImmutableMap<String, String> tags;
 
-    @JsonCreator
-    public DefaultResultSerializer(@JsonProperty("typeNames") List<String> typeNames,
-                                   @JsonProperty("booleanAsNumber") boolean booleanAsNumber,
-                                   @JsonProperty("rootPrefix") String rootPrefix,
-                                   @JsonProperty("tags") Map<String, String> tags) {
-        this.jsonFactory = new JsonFactory();
-        this.typeNames = ImmutableList.copyOf(typeNames);
-        this.booleanAsNumber = booleanAsNumber;
-        this.rootPrefix = rootPrefix;
-        this.tags = ImmutableMap.copyOf(tags);
-    }
+	@JsonCreator
+	public DefaultResultSerializer(@JsonProperty("typeNames") List<String> typeNames,
+								   @JsonProperty("booleanAsNumber") boolean booleanAsNumber,
+								   @JsonProperty("rootPrefix") String rootPrefix,
+								   @JsonProperty("tags") Map<String, String> tags) {
+		this.jsonFactory = new JsonFactory();
+		this.typeNames = ImmutableList.copyOf(typeNames);
+		this.booleanAsNumber = booleanAsNumber;
+		this.rootPrefix = rootPrefix;
+		this.tags = ImmutableMap.copyOf(tags);
+	}
 
-    @Nonnull
-    @Override
-    public Collection<String> serialize(Server server, Query query, Result result) throws IOException {
-        log.debug("Query result: [{}]", result);
-        Map<String, Object> resultValues = result.getValues();
-        List<String> messages = new ArrayList<>();
-        for (Map.Entry<String, Object> values : resultValues.entrySet()) {
-            Object value = values.getValue();
-            if (isNumeric(value)) {
-                messages.add(createJsonMessage(server, query, result, values, value));
-            } else {
-                log.warn("Unable to submit non-numeric value to Kafka: [{}] from result [{}]", value, result);
-            }
-        }
-        return messages;
-    }
+	@Nonnull
+	@Override
+	public Collection<String> serialize(Server server, Query query, Result result) throws IOException {
+		log.debug("Query result: [{}]", result);
+		Map<String, Object> resultValues = result.getValues();
+		List<String> messages = new ArrayList<>();
+		for (Map.Entry<String, Object> values : resultValues.entrySet()) {
+			Object value = values.getValue();
+			if (isNumeric(value)) {
+				messages.add(createJsonMessage(server, query, result, values, value));
+			} else {
+				log.warn("Unable to submit non-numeric value to Kafka: [{}] from result [{}]", value, result);
+			}
+		}
+		return messages;
+	}
 
-    private String createJsonMessage(Server server, Query query, Result result, Map.Entry<String, Object> values, Object value) throws IOException  {
-        String keyString = getKeyString(server, query, result, values, typeNames, this.rootPrefix);
-        String cleanKeyString = keyString.replaceAll("[()]", "_");
+	private String createJsonMessage(Server server, Query query, Result result, Map.Entry<String, Object> values, Object value) throws IOException {
+		String keyString = getKeyString(server, query, result, values, typeNames, this.rootPrefix);
+		String cleanKeyString = keyString.replaceAll("[()]", "_");
 
-        try (
-                ByteArrayOutputStream out = new ByteArrayOutputStream();
-                JsonGenerator generator = jsonFactory.createGenerator(out, UTF8)
-        ){
-            generator.writeStartObject();
-            generator.writeStringField("keyspace", cleanKeyString);
-            generator.writeStringField("value", value.toString());
-            generator.writeNumberField("timestamp", result.getEpoch() / 1000);
-            generator.writeObjectFieldStart("tags");
+		try (
+				ByteArrayOutputStream out = new ByteArrayOutputStream();
+				JsonGenerator generator = jsonFactory.createGenerator(out, UTF8)
+		) {
+			generator.writeStartObject();
+			generator.writeStringField("keyspace", cleanKeyString);
+			generator.writeStringField("value", value.toString());
+			generator.writeNumberField("timestamp", result.getEpoch() / 1000);
+			generator.writeObjectFieldStart("tags");
 
-            for (Map.Entry<String, String> tag : this.tags.entrySet()) {
-                generator.writeStringField(tag.getKey(), tag.getValue());
-            }
+			for (Map.Entry<String, String> tag : this.tags.entrySet()) {
+				generator.writeStringField(tag.getKey(), tag.getValue());
+			}
 
-            generator.writeEndObject();
-            generator.writeEndObject();
-            generator.close();
-            return out.toString("UTF-8");
-        }
-    }
+			generator.writeEndObject();
+			generator.writeEndObject();
+			generator.close();
+			return out.toString("UTF-8");
+		}
+	}
 }
--- a/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/DefaultResultSerializer.java
+++ b/jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/DefaultResultSerializer.java
@@ -31,6 +31,8 @@ import com.google.common.collect.ImmutableMap;
 import com.googlecode.jmxtrans.model.Query;
 import com.googlecode.jmxtrans.model.Result;
 import com.googlecode.jmxtrans.model.Server;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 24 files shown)


jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter2.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public KafkaWriter2(@Nonnull ObjectMapper objectMapper, @Nonnull Map<String, Object> producerConfig, @Nonnull String topic) {
+        this.objectMapper = objectMapper;
+        this.producerConfig = producerConfig;
+        this.topic = topic;
+    }
+
+    @Override
+    public void start() throws LifecycleException {
+        producer = new KafkaProducer<String, String>(producerConfig);
```
gehel: You should move all object construction to the Factory.

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterFactory.java
```diff
@@ -0,0 +1,63 @@
...
+import lombok.Getter;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.serialization.StringSerializer;
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+@EqualsAndHashCode(exclude = {"objectMapper", "producerConfig"})
```
gehel: You should most probably not exclude `producerConfig` from `@EqualsAndHashCode`. To writers with different configuration are not behaving in the same way.

jmxtrans-output/jmxtrans-output-kafka/src/test/java/com/googlecode/jmxtrans/model/output/kafka/EmbeddedZookeeper.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Properties;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Embedded Zookeeper server for integration testing
+ */
+public class EmbeddedZookeeper {
```
gehel: This looks very much like something that should be implemented as a junit `@Rule`

jmxtrans-output/jmxtrans-output-kafka/src/test/java/com/googlecode/jmxtrans/model/output/kafka/EmbeddedKafka.java
```diff
@@ -0,0 +1,67 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+
+import static com.googlecode.jmxtrans.model.output.kafka.EmbeddedZookeeper.getResourceAsProperties;
+
+public class EmbeddedKafka {
```
gehel: This looks very much like something that should be implemented as a junit `@Rule`

gehel:
Yes, the split Writer / Factory is very much going in the right direction! A few things to improve, but you're doing good!

jmxtrans-output/jmxtrans-output-kafka/src/test/java/com/googlecode/jmxtrans/model/output/kafka/EmbeddedZookeeper.java
gquintana: I thought about creating a rule but:
- As Kafka depends on Zookeeper, Zookeeper must be started before Kafka and stopped after Kafka, I thinks it's clearer to tell it.
- Both Kafka and Zookeeper depend on a data folder I create with the TemporaryFolder rule

I will try if a RuleChain may do the job

gquintana:
I am not very satisfied with the way Query Result gets serialized into a JSON string, because this something that should be made configurable and reusable (the Elastic Writer probably has the same requirement)
1. how to name fields and format their values (JSON schema) ?
2. one message per query / one message per attribute ?
3. adding contextual information (tags, environment, jmxtrans host...)?
4. how to make this pluggable (and reusable)?

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterFactory.java
gquintana: Comparing maps looks dangerous to me...

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriter2.java
gquintana: I don't how known how the lifecycle of writers is done, but once closed, the Kafka producer ca not be restarted. Exception that I can move it in the writer constructor.

However, moving the creation of the KafkaProducer outside of the writer (in the writer factory) means exposing (opening) the internals of the writer to the outside. It doesn't look clean to me.

Yet I admit I do not understand the intent behind splitting writer/writer factory.

gehel: The OutputWriters are not meant to be restarted, so moving the initialization of the Kafka producer out of the writer is no problem on that ground.

The intent of splitting writer / factory has multiple facets (I'm happy to do a Google hagouts about it, if you want to chat a bit more about the details). Here are some of them:

* **splitting parsing / construction and execution:** Output writers were becoming messy, with too many responsibilities. Json parsing in the same class as the writer itself was cluttering the code, and make it harder to thing just about the execution of the writer or just about the parsing.
* **simpler lifecycle:** I really don't like that `start()` method in the `OutputWriter`. An object should be ready to be used once constructed. But since we usually have non trivial construction work, this is not well suited to be done in a constructor. Moving it to a factory makes the writer itself much simpler (at some poitn, I want to remove that `start()`method).
* **composability:** A lot of the writers can share a lot of components. Creating writers by composing existing classes is much easier if it is not done in the constructor ([prefer composition over inheritance applies here](https://en.wikipedia.org/wiki/Composition_over_inheritance)). As an example, the [`GraphiteWriterFactory`](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/GraphiteWriterFactory.java#L104-L128) can create both TCP or UDP based writers. And the UDP / TCP part is shared by multiple writers.

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterFactory.java
```diff
@@ -0,0 +1,89 @@
...
+    @Getter
+    private final ResultSerializer resultSerializer;
+
+    private static final String STRING_SERIALIZER_CLASS = StringSerializer.class.getName();
+
+    @JsonCreator
+    public KafkaWriterFactory(
+            @JsonProperty("producerConfig") @Nonnull Map<String, Object> producerConfig,
+            @JsonProperty("topic") @Nonnull String topic,
+            @JsonProperty("resultSerializer") ResultSerializer resultSerializer) {
```
gehel: I'd prefer to not expose ResultSerializer as a JsonProperty, exposing only simple key values in the Factories and choosing the correct class in the implementation. This will make it easier to refactor the code down the road, while preserving the original configuration interface.

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/DefaultResultSerializer.java
```diff
@@ -0,0 +1,113 @@
...
+        this.jsonFactory = new JsonFactory();
+        this.typeNames = ImmutableList.copyOf(typeNames);
+        this.booleanAsNumber = booleanAsNumber;
+        this.rootPrefix = rootPrefix;
+        this.tags = ImmutableMap.copyOf(tags);
+    }
+
+    @Nonnull
+    @Override
+    public Collection<String> serialize(Server server, Query query, Result result) throws IOException {
```
gehel: Introducing a generic Json writer is a great idea! We should make it a bit more generic, by passing a json Generator as a method argument here, and letting the caller create that object. This would allow to inject a generator that is directly based on an output stream and getting rid of this intermediate String (for the generic case at least, it might not be useful for Kafka itself).

This comment might be a bit obscure, I'll try to find some time to propose actual code, which might be easier to understand. And again, if you want to have a chat on all that, just ping me.

gquintana:
Can you help in fixing the "missing license" error in the build? I added the header, but it doesn't detect it.

About licensing, I copied/pasted/tuned the zookeeper/kafka config files which are under Apache License, is it a problem?

gehel:
The easiest way to add licenses is to use `mvn license:format` which shoudl add the proper licenses to all files.

As far as I understand, Apache license is fine to include. You probably want to add an exception to the maven license plugin so that it does not try to rewrite those headers.

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterFactory.java
gquintana: Exposing the resultserializer allows to easily change the serialization strategy (at the moment Default/Detail, but could also be custom one).

Yet I agree I can make it bit more backward compatible.

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/DefaultResultSerializer.java
gquintana: I thought about passing a java.io.Writer|InputStream instead of returning Strings.
But it makes converting one single Result into a collection of JSONs (one per field in case of hash) more complicated.
https://github.com/jmxtrans/jmxtrans/pull/586/files#diff-71873b59a40470a5ab0bdf1fe09b5131R96

jmxtrans-output/jmxtrans-output-kafka/src/main/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterFactory.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.serialization.StringSerializer;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+@EqualsAndHashCode(exclude = {"objectMapper"})
```
gehel: Since the ResultSerializer is used in equals / hashCode, the whole class hierarchy needs to implement a meaningful euqals / hashCode (and sadly there is no good way to ensure that in Java).

jmxtrans-output/jmxtrans-output-kafka/src/test/resources/kafka.properties
```diff
@@ -0,0 +1,128 @@
+#
```
gehel: Didn't you say that config files are Apache licensed? Though to be honest, I'm not sure there is enough creation in this to be copyrightable...

gquintana:
I am having hard times finding which code needs testing...
```
 project failed coverage check. Branch coverage rate of 40.8% is below 50.0%
```

gehel:
Not sure why cobertura reports only 48% branch coverage... something to investigate... In the meantime, the coverage looks more than adequate. You can just lower the constraint in the [pom.xml](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-kafka/pom.xml#L40).

jmxtrans-output/jmxtrans-output-kafka/src/test/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterIT.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.offset;
+import static org.assertj.core.api.Assertions.within;
+
+public class KafkaWriterIT {
```
gehel: This should be tagged with `@Category({IntegrationTest.class, RequiresIO.class})` so that pitest does not try to run it.

jmxtrans-output/jmxtrans-output-kafka/src/test/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterTests.java
```diff
@@ -45,42 +46,45 @@
...
+	public void
 	kafkaWriterNotNull() throws Exception {
 		assertThat(getTestKafkaWriter()).isNotNull();
 	}
 
-	@Test public void
+	@Test
+	public void
 	messagesAreSentToKafka() throws Exception {
 		KafkaWriter writer = getTestKafkaWriter();
```
gehel: You should probably ensure that the KafkaWriter is closed properly (try-with-resources)

jmxtrans-output/jmxtrans-output-kafka/src/test/java/com/googlecode/jmxtrans/model/output/kafka/KafkaWriterTests.java
```diff
@@ -45,42 +46,46 @@
...
+	@Mock
+	private Producer<String, String> producer;
+	@Captor
+	private ArgumentCaptor<ProducerRecord<String, String>> messageCaptor;
 
-	@Test public void
+	@Test
+	public void
 	kafkaWriterNotNull() throws Exception {
 		assertThat(getTestKafkaWriter()).isNotNull();
```
gehel: Still missing one writer that should be closed :)

gehel:
Note: you can run the mutation tests locally with `mvn verify withMutationTests`. This will produce a report that will show where your tests might need to be improved. Or you can just ignore them and lower the required mutation score in the module's [pom.xml](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-kafka/pom.xml#L39). I have obviously a bias toward improving the tests, but it might or might not make sense in this case. I'll have a look at the report and see if there is anything that obviously needs to be improved.

gehel:
It looks like the pitest report has at least a few suggestions which make sense. A few more assertions would be nice... Let me know if you are lost!

gquintana:
I got 61% in mutation testing. Is it enough? Where should I add tests?

gehel:
Enough test is when everything that makes sense to test is tested, not some arbitrary number. I do add a minimum threshold in each module (that I try to update to reflect the current coverage), so that when coverage drops, I make a real decision about dropping coverage. I have no problem with dropping those limits if you feel that the coverage is adequate.

That being said, a few things that might make sense to test (if you really want to, not absolutely necessary):

* the message is valid json: at the moment, [dropping the closing of json object](https://github.com/jmxtrans/jmxtrans/pull/586/files#diff-b92987e6052610899184caf062454fd0R114) does not fail the test.
* there isn't an assertion that validates that each field of KResult is set correctly (this one is really minor)
* there is no assertion to validate that KafkaWriter / KafkaWriter2 closes the KafkaProducer. Testing this would probably require to refactor KafkaWriter to have the KafkaProducer injected. I know you disagree with this, but I do find injecting it cleaner in any case.

Again, I'm perfectly fine with the current coverage, feel free to lower the check to 60%.

It looks like this PR is really close to being ready to be merged. It has grown quite a bit, I'll need some time to review. Thanks a lot for your patience!

gquintana:
I fully agree that lowering code coverage is not the solution, I also prefer adding tests where needed.

I'll add tests to check that producer is closed (I can already inject a mock producer) and all fields get filled.

Testing whether JSON is closed is mostly done: in the KafkaWriterTest, I replaced "string contains " assertions by something parsing the JSON. But Jackson is smart, either it autocloses curly braces at the time of writing, or it can parse unfinished JSON at the time of reading. Maybe I should use a  simpler JSON library at test time.

Thanks for your advice and code reviews.

gquintana:
In https://github.com/jmxtrans/jmxtrans/pull/586/files#diff-b92987e6052610899184caf062454fd0R114
the `generator.writeEndObject()` lines are indeed useless...

Improved mutation testing to 63% and branch coverage to 57%

gehel:
If you feel that the test coverage is good enough, feel free to lower the requirements to make the build pass!

gquintana:
@gehel The build passes now, tell me what shall be done.

Issue #371, opened 2015-12-08
InfluxDd writes _jmx_port as a field instead of a tag


vpernin:
InfluxDd writes _jmx_port as a field instead of a tag.
So it becomes impossible to group by _jmx_port

willmpls:
[PR #377] (https://github.com/jmxtrans/jmxtrans/pull/377) has been submitted and addresses this one.  _jmx_port removed as a field and added as a tag.

PR #591, opened 2017-07-19
InfluxDb output : allowing to report JMX port as a tag

hightea:
Add a property "reportJmxPortAsTag" to InfluxDbWriter allowing users to specify if the JMX port is written as a field (current and default behavior) or as a tag (cf. issue #371).

### DIFF
--- a/jmxtrans-output/jmxtrans-output-influxdb/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterTests.java
+++ b/jmxtrans-output/jmxtrans-output-influxdb/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterTests.java
@@ -33,8 +33,6 @@ import com.googlecode.jmxtrans.model.Result;
 import com.googlecode.jmxtrans.model.ResultAttribute;
 import com.googlecode.jmxtrans.model.ResultAttributes;
 import com.googlecode.jmxtrans.util.ProcessConfigUtils;
-import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang.WordUtils;
 import org.influxdb.InfluxDB;
 import org.influxdb.InfluxDB.ConsistencyLevel;
 import org.influxdb.dto.BatchPoints;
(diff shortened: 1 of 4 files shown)


jmxtrans-output/jmxtrans-output-influxdb/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterTests.java
```diff
@@ -117,7 +119,13 @@ public void pointsAreWrittenToInfluxDb() throws Exception {
 		assertThat(points).hasSize(1);
 
 		Point point = points.get(0);
-		assertThat(point.lineProtocol()).startsWith(lineProtocol);
+		String pointLineProtocol = point.lineProtocol();
```
gehel: I find this added `pointLineProtocol` variable does not add readability. I much prefer inlining it.

jmxtrans-output/jmxtrans-output-influxdb/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterTests.java
```diff
@@ -117,7 +119,13 @@ public void pointsAreWrittenToInfluxDb() throws Exception {
...
+		assertThat(pointLineProtocol).startsWith(lineProtocol);
+		// JMX port is written as field by default :
+		// cf Point.lineProtocol() : there is a space after the tags and one after the fields
+		String[] split = pointLineProtocol.split(" ");
```
gehel: Could you please find a better name for `split`?

jmxtrans-output/jmxtrans-output-influxdb/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterTests.java
```diff
@@ -117,7 +119,13 @@ public void pointsAreWrittenToInfluxDb() throws Exception {
...
+		assertThat(split).hasSize(3);
+		assertThat(split[1]).contains(JMX_PORT_KEY + "=" + DEFAULT_PORT);
```
gehel: Does it matter that JMX_PORT_KEY is at position `1`? Or should we just test that there is an entry for it?

jmxtrans-output/jmxtrans-output-influxdb/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterTests.java
```diff
@@ -215,6 +223,27 @@ public void databaseIsNotCreated() throws Exception {
 	}
 
 	@Test
+	public void jmxPortIsReportedAsTag() throws Exception {
+		InfluxDbWriter writer = getTestInfluxDbWriterWithReportJmxPortAsTag();
```
gehel: A good section of this method is duplicated from `pointsAreWrittenToInfluxDb()`. There are probably two issues here:

1) the existing tests don't check a single thing, but multiple (which means that you are actually duplicating the tests here, which is not needed). The existing test shoudl be refactor so that the common behaviour is tested only once, and the specifics are tested in different tests.
2) some common behaviour (setting up the mocks and captors) are identical and should be extracted in their own methods.

gehel:
This looks good! A few minor comments inline, if you could have a look and fix them...

jmxtrans-output/jmxtrans-output-influxdb/src/test/java/com/googlecode/jmxtrans/model/output/InfluxDbWriterTests.java
hightea: Yes the position matters : if JMX_PORT_KEY is in the first token (split[0]), that means that it is reported as a tag, whereas if it is in the second token (split[1]), it is reported as a field.
However I'll comment this a little better.

hightea:
I'll rework the tests and push the updates. Thanks for your review.

hightea:
I've rebased this PR, but Appveyor seems to be unhappy. Is there a way to re-run it ? cc @gquintana

PR #606, opened 2017-09-14
Refactor JmxResultProcessor

gquintana:
I tried to make the JmxResultProcessor cleaner by:
* Creating and internal Result value walker/accumulator instead of passing the List.Builder to each method
* Replacing TabularDataSupport/CompositeDataSupport by their interface counterpart TabularData/CompositeData
* Extracted each if block from the original getResult method in its method. Split the getResult method in small pieces.
* Factorised the Result r = getNewResultObject(...) + accumulator.add(r) pattern in a addNew method
It behaves like the original version

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
@@ -22,6 +22,7 @@
  */
 package com.googlecode.jmxtrans.model;
 
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
@@ -29,9 +30,8 @@ import javax.management.Attribute;
 import javax.management.ObjectInstance;
 import javax.management.ObjectName;
 import javax.management.openmbean.CompositeData;
-import javax.management.openmbean.CompositeDataSupport;
 import javax.management.openmbean.CompositeType;
-import javax.management.openmbean.TabularDataSupport;
+import javax.management.openmbean.TabularData;
 import java.lang.reflect.Array;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
```diff
@@ -58,132 +58,137 @@ public JmxResultProcessor(Query query, ObjectInstance objectInstance, List<Attri
...
+			for (String key : keys) {
+				Object value = cds.get(key);
+				if (value instanceof TabularData) {
+					TabularData tds = (TabularData) value;
+					add(attributeName + "." + key, tds);
+					// continue because we added tabular contents within above, but need primitives at this level
+				} else if (value instanceof CompositeData) {
+					// now recursively go through everything.
+					CompositeData cds2 = (CompositeData) value;
+					add(attributeName, cds2);
```
gquintana: This part looks very strange:
* TabularData: add result with key "attributeName + data key"
* CompositeData: only add a result with the first element with key "attributeName"
* Simple type (I guess): add a result of type map
Personally, I would remove this "if tabular ... else if composite ... else ..." block and call the add(name, object) with name = name + key. This also means I would change the type of the "values" field in Result class from Map to Object, meaning each Result would contain a single and simple value.

What do you think?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
```diff
@@ -58,132 +58,137 @@ public JmxResultProcessor(Query query, ObjectInstance objectInstance, List<Attri
...
+			} else if (value instanceof CompositeData[]) {
+				for (CompositeData cd : (CompositeData[]) value) {
+					add(name, cd);
+				}
+			} else if (value instanceof ObjectName[]) {
+				add(name, (ObjectName[]) value);
+			} else if (value.getClass().isArray()) {
+				// OMFG: this is nutty. some of the items in the array can be
+				// primitive! great interview question!
+				Map<String, Object> values = newHashMap();
```
gquintana: * Array of ObjectName: add one Result with a special map
* Array of CompositeData: add Results for each CompositeData
* Array of anything else: add one Result with map similar to the array

gquintana:
@gehel Can you give me feedback on this refactoring?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
gehel: I think that each Result being a single object makes a lot of sense. If we manage to get each Result to be a single scalar value, this allow to simplify the OutputWriter quite a bit (I do suspect that a lot of them are broken and don't handle collections well).

Note for later: I think it would also make sense to abstract the metric name / path, instead of relying on strings. But that would also require modifying all existing OutputWriters...

gehel:
Thanks for the ping! I think this PR definitely simplifies things! Thanks for digging into this mess! Feel free to merge it (I have fairly limited time right now, sorry for the delay).

PR #611, opened 2017-10-03
Update NagiosWriter.java

WinterUnicorn:
Fix NullPointerException when check filters/thresholds in constructor.


gehel:
Thanks for the fix!

PR #614, opened 2017-10-10
Make Result contain a single value

gquintana:
* Result: Replace Map<String,Object> values by List<String> valuePath + Object value (where value is supposed to be simple)
* JmxResultProcessor: recursively walk in the CompositeData/TabularData/Array/Collection and add a different result with valuePath containing the path 
* All :Writers
  * Remove iteration over the Map<String,Object> values 
  * Add List<String> valuePath to metric key
  * Replacer result.getAttributeName().equals(valuesEntry.getKey()) by result.getValuePath().isEmpty()

TODO
* Improve code style
* Improve code coverage


gquintana:
This is a big change which makes code simpler but may introduce changes in how metrics are named/identified. I may have broken integration with monitoring tools, with existing dashboards, with waiting PR.

Tell me if you like this PR (I'll finish it) or not (I'll drop it).

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java
```diff
@@ -52,18 +51,20 @@
 	@Getter private final String className;
 	@Getter private final String objDomain;
 	@Getter private final String typeName;
-	@Nonnull @Getter private final ImmutableMap<String, Object> values;
+	@Getter private final ImmutableList<String> valuePath;
```
gehel: I almost think that valuePath should be its own type, completely hiding the implementation. That being said, you have read all that messy code more recently than I have, so feel free to have a different opinion!

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java
```diff
@@ -52,18 +51,20 @@
...
+	@Nonnull @Getter private final Object value;
```
gehel: While the values Map was ensured to be non null, I'm pretty sure that a single value can be null...

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/ResultFixtures.java
```diff
@@ -36,7 +35,8 @@ public static Result booleanTrueResult() {
 				"ObjectDomainName",
 				"VerboseMemory",
 				"type=Memory",
-				ImmutableMap.<String, Object>of("Verbose", true));
+				ImmutableList.<String>of(),
```
gehel: It would be nice if the valuePath was coherent with the rest of the parameters (I'm sure you intended to do the cleanup later...)

gehel:
Yeah, that is a slightly scary change! But it does simplifies things a lot! Thanks for taking the time! And yes, this cleanup is really welcomed!

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/naming/KeyUtils.java
```diff
@@ -22,12 +22,12 @@
...
 
+import com.google.common.base.Joiner;
 import com.googlecode.jmxtrans.model.Query;
 import com.googlecode.jmxtrans.model.Result;
 import com.googlecode.jmxtrans.model.Server;
 
 import java.util.List;
-import java.util.Map;
 
 public final class KeyUtils {
```
gehel: Note for later: If we get a clean abstraction on metrics names (and you are starting that way with introducing valuePath), then we can start converting this utility class to a naming strategy. The "." separated string is what a lot of OutputWriter need, but it is far from universal. And having this in a static class makes it harder to compose...

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java
gquintana: What do you imagine in this specific type?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java
gquintana: The JmxResultProcessor throughs away null values so I left @Nonnull but that may not be good idea.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/ResultFixtures.java
gquintana: What do you mean by "coherent with the rest of the parameters"

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/naming/KeyUtils.java
gquintana: I thought about adding a getValuePathAsString method in Result class, because at the moment almost all writer implementations use this valuePath as dot delimited String. At the end I placed this code in KeyUtils class.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/ResultFixtures.java
gehel: Oh, my bad. In this case it is actually expected to have an empty valuePath.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java
gehel: Actually, I'm imagining something more generic than just `valuePath` (I was reading all this a bit too fast). If I understand correctly, `valuePath` is the path specific to the value, in the context of the specific bean. I was imagining a more generic abstraction, which would care about the full path (or name, or context, or coordinate, not sure about the name) of any value. This more generic object would abstract both the `valuePath` and what is inferred from the `className`, `objName`, `typeName`, ... But this probably belong to another change.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java
gehel: If we do throw away null values, then `@NonNull` is fine. Though I now wonder if it is a good thing or not...

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/naming/KeyUtils.java
gehel: Yep, `KeyUtils` look like the right place to me too. At least until we refactor this a bit more.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Result.java
gquintana: At the moment, valuePath contains the path from the MBean attribute to the simple (numeric) value:
* Empty when the attribute is already a simple value (java.lang:type=Threading ThreadCount [])
* Key when the attribute is Map or a CompositeData (java.lang:type=Memory HeapMemoryUsage [used])
* Numerical index when the attribute is Array or Collection 
* Row Key when the attribute is TabularData (java.lang:type=Runtime SystemProperties [java.version, value] or [java.version, key])

gehel: yep, that make sense as a first step!

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/ResultFixtures.java
gquintana: Yes one of the changes related to removing the Map is that in case of simple value we don't have to choose an arbitrary Map key.

gquintana:
This PR is hard to maintain in sync with master because changes are scattered ovel the whole code base.

Prior to merging on master, a release should be done on master to mark a last release before major API change.

gquintana:
Concerning #617 you can see that the test now returns a empty list: null values are properly skipped.

gehel:
I launched a release (https://travis-ci.org/jmxtrans/jmxtrans/builds/302607280). Feel free to merge this PR once that's done.

gehel:
Release is completed.

Issue #615, opened 2017-10-11
Infinity causes NumberFormatException in CPrecisionValueTransformer


salbanese:
When any of the _INFINITY values are passed to CPercisionValueTransformer it throws a NumberFormatException.

unit test showing the issue [here](https://github.com/salbanese/jmxtrans/compare/master...handle-infinity)

This is very similar to #529 and I'm happy to put in a similar fix if there are no objections.

PR #616, opened 2017-10-11
Handle infinity in CPercisionValueTransformer

salbanese:
fix for #615 

CPercisionValueTransformer now returns null if input is infinite


gehel:
Thanks for the fix!

Issue #626, opened 2017-12-04
Collecting metrics stops when at least one node is down


kipwoker:
My collegue wrote (https://gitter.im/jmxtrans/jmxtrans?at=58e4fdaaad849bcf425405a5)

> Hi! I’m observing a strange side effect when using jmxtrans (build 260) to collect metrics from Cassandra nodes. When at least one node is down, I observe incorrect or missing datapoints on all other metrics coming from the same instance of jmxtrans.
What am I doing wrong?

We've tried to fix this. But I have no idea what's wrong.
There are my investigation results. May be it helps to get answer.

First of all we've covered by logs next methods

- Server.execute
- JmxUtils.processServer

After that we're replayed this scenario at production. One cassandra node was closed by firewall to test instance of jmxtrans.

What actually happened.

Logs from JmxUtils.processServer when we closed node by firewall and our thread pool limit set to 10:
1. ThreadPool ActiveCount=10 | QueueSize=10. All items in queue for node which is down.
2. After one minute we have no metrics from another nodes.
3. After three minutes ThreadPool has stats ActiveCount=10 | QueueSize=2530 and no errors at log.
4. Rollback firewall block.
5. After few seconds app wrote to log some errors about connection timeout. Like this:
```
[01 Dec 2017 14:16:14] [jmxtrans-query-0] 379842 ERROR (com.googlecode.jmxtrans.jmx.ProcessQueryThread:57) - Error executing query Query(objectName=org.apache.cassandra.metrics:type=ColumnFamily,keyspace=*,scope=*,name=MemtableSwitchCount, keys=[], attr=[Value], typeNames=[], resultAlias=ColumnFamily, useObjDomainAsKey=false, allowDottedKeys=true, useAllTypeNames=false, outputWriterInstances=[GraphiteWriter(pool=org.apache.commons.pool.impl.GenericKeyedObjectPool@29f9efa0, rootPrefix=DevOpsTest.Cassandra.WebCluster, address=graphite-relay/7.7.7.7:2003, onlyOnceLogger=com.googlecode.jmxtrans.util.OnlyOnceLogger@457b9597)]) on server Server(pid=null, host=cassandra-134, port=8899, url=service:jmx:rmi:///jndi/rmi://cassandra-134:8899/jmxrmi, cronExpression=null, numQueryThreads=20)
java.rmi.ConnectException: Connection refused to host: 192.168.191.2; nested exception is: 
[...]

PR #627, opened 2017-12-07
JMXConnector works slowly when server unavailable

kipwoker:
JMXConnector works slowly when server unavailable and executor thread pool fills up quickly. This changes can reduce awaiting duration.

This fixes refers to https://github.com/jmxtrans/jmxtrans/issues/626

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
@@ -24,6 +24,8 @@ package com.googlecode.jmxtrans.connections;
 
 import lombok.Getter;
 import lombok.ToString;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -32,20 +34,45 @@ import javax.management.MBeanServerConnection;
 import javax.management.remote.JMXConnector;
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 @ToString
 @ThreadSafe
 public class JMXConnection implements Closeable {
 	@Nullable private final JMXConnector connector;
 	@Nonnull @Getter private final MBeanServerConnection mBeanServerConnection;
+	@Nonnull private final ExecutorService executor;
+	private boolean markedAsDestroyed;
+	private static final Logger logger = LoggerFactory.getLogger(JMXConnection.class);
 
 	public JMXConnection(@Nullable JMXConnector connector, @Nonnull MBeanServerConnection mBeanServerConnection) {
 		this.connector = connector;
 		this.mBeanServerConnection = mBeanServerConnection;
+		this.markedAsDestroyed = false;
+
+		executor = Executors.newSingleThreadExecutor();
+	}
+
+	public boolean isAlive(){
+		return !markedAsDestroyed;
 	}
 
 	@Override
 	public void close() throws IOException {
-		if (connector != null) connector.close();
+		markedAsDestroyed = true;
+		if (connector != null) {
+			executor.submit(new Runnable() {
+				public void run() {
+					if (connector != null) {
+						try {
+							connector.close();
+						} catch (IOException e) {
+							logger.error("Error occurred during close connection {}", this, e);
+						}
+					}
+				}
+			});
+		}
 	}
 }
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
@@ -34,15 +34,12 @@ import javax.management.MBeanServerConnection;
 import javax.management.remote.JMXConnector;
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 
 @ToString
 @ThreadSafe
 public class JMXConnection implements Closeable {
 	@Nullable private final JMXConnector connector;
 	@Nonnull @Getter private final MBeanServerConnection mBeanServerConnection;
-	@Nonnull private final ExecutorService executor;
 	private boolean markedAsDestroyed;
 	private static final Logger logger = LoggerFactory.getLogger(JMXConnection.class);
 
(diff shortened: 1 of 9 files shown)


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
```diff
@@ -32,20 +34,45 @@
...
+	public boolean isAlive(){
+		return !markedAsDestroyed;
 	}
 
 	@Override
 	public void close() throws IOException {
-		if (connector != null) connector.close();
+		markedAsDestroyed = true;
+		if (connector != null) {
+			executor.submit(new Runnable() {
```
gehel: Why delegate the closing to an executor?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/guice/JmxTransModule.java
```diff
@@ -84,11 +84,11 @@ public JmxTransModule(@Nonnull JmxTransConfiguration configuration) {
...
 	protected void configure() {
 		bind(new TypeLiteral<GenericKeyedObjectPool<InetSocketAddress, Socket>>(){})
-				.toInstance(getObjectPool(new SocketFactory(), SocketFactory.class.getSimpleName()));
+				.toInstance(getObjectPool(new SocketFactory(), SocketFactory.class.getSimpleName(), 0));
 		bind(new TypeLiteral<GenericKeyedObjectPool<SocketAddress, DatagramSocket>>(){})
-				.toInstance(getObjectPool(new DatagramSocketFactory(), DatagramSocketFactory.class.getSimpleName()));
+				.toInstance(getObjectPool(new DatagramSocketFactory(), DatagramSocketFactory.class.getSimpleName(), 0));
 		bind(new TypeLiteral<KeyedObjectPool<JmxConnectionProvider, JMXConnection>>(){}).annotatedWith(Names.named("mbeanPool"))
-				.toInstance(getObjectPool(new MBeanServerConnectionFactory(), MBeanServerConnectionFactory.class.getSimpleName()));
+				.toInstance(getObjectPool(new MBeanServerConnectionFactory(), MBeanServerConnectionFactory.class.getSimpleName(), 20000));
```
gehel: It would be great to expose this as a cli argument (but not absolutely required). We need to introduce a config file instead of passing all options via cli, the current argument processing is messy and overly complex.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
kipwoker: I wrote it in commit message: when you call close it tries to notify server about it, so if your server is down you will wait until connection timed out.
This is just for fast release object from pool => fast release thread from queryExecutor thread pool.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/guice/JmxTransModule.java
kipwoker: It's good idea to make it parameter. I wrote your thoughts about config file at https://github.com/jmxtrans/jmxtrans/pull/566 and agree with them, but it seems like separate task.

So I want to extract it like cli parameter. Is it okay if I will do another pull request for this?

gehel: sure, all good!

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
gehel: Ok, that make sense. Adding those thoughts as an inline comment would be nice.

kipwoker: I've added comment. Check it out.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
```diff
@@ -32,20 +34,49 @@
...
+	@Nonnull private final ExecutorService executor;
+	private boolean markedAsDestroyed;
+	private static final Logger logger = LoggerFactory.getLogger(JMXConnection.class);
 
 	public JMXConnection(@Nullable JMXConnector connector, @Nonnull MBeanServerConnection mBeanServerConnection) {
 		this.connector = connector;
 		this.mBeanServerConnection = mBeanServerConnection;
+		this.markedAsDestroyed = false;
+
+		executor = Executors.newSingleThreadExecutor();
```
gehel: We should probably extract that executor at least as high as `MBeanServerConnectionFactory` and share it between connections. Keeping a single thread executor is probably fine for this case (if we have enough servers in trouble to saturate the single thread executor, we probably want to apply some back pressure to the connection pool anyway, we can tune that in the future if need be).

This would also allow the use of a sameThreadExecutor in the test, slightly simpler.

kipwoker: Yeah, I thought about it. I imagined this situation. It will calls rarely.
But your sentece sounds good. I will try to fix it.

kipwoker:
@gehel I moved executor to factory. Sorry for merge commit my branch was outdatted.

gehel:
I do have a preference to rebase pull requests instead of merging them. But don't worry about it!

kipwoker:
Okay. Thank you. Next time rebase only. :)

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxTransRMIClientSocketFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.IOException;
+import java.net.Socket;
+import java.rmi.server.RMIClientSocketFactory;
+import java.rmi.server.RMISocketFactory;
+
+public class JmxTransRMIClientSocketFactory implements RMIClientSocketFactory {
+	private final int timeoutMillis;
+	private RMIClientSocketFactory rmiClientSocketFactory;
+
+	public JmxTransRMIClientSocketFactory(int timeoutMillis, boolean ssl){
```
gquintana: I am surprised this doesn't break the SSL connection test because in `javax.management.remote.rmi.RMIConnector#checkStub` there is this (ugly) piece of code:
```
        // Check RMIClientSocketFactory in stub is from the expected class
        // "javax.rmi.ssl.SslRMIClientSocketFactory".
        //
        LiveRef liveRef = ((UnicastRef2)ref).getLiveRef();
        RMIClientSocketFactory csf = liveRef.getClientSocketFactory();
        if (csf == null || csf.getClass() != SslRMIClientSocketFactory.class)
            throw new SecurityException(
                    "Expecting a " + SslRMIClientSocketFactory.class.getName() +
                    " RMI client socket factory in stub!");
```
I stumbled uppon it a long time ago, I didn't check it again though.

kipwoker: Wow. I didnt' know about this. But your right it seems strange.

PR #629, opened 2017-12-13
Separate thread pool executors

kipwoker:
As I mentioned at https://github.com/jmxtrans/jmxtrans/issues/626 I found trouble behavior with common thread pool. For backward compatibility I implemented it though another strategy. I've tried it on production and it works. Dead server didn't affect another servers.

In future it can be extended. Server config may contain executor settings for each server.

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
@@ -31,6 +31,7 @@ import com.googlecode.jmxtrans.classloader.ClassLoaderEnricher;
 import com.googlecode.jmxtrans.cli.JCommanderArgumentParser;
 import com.googlecode.jmxtrans.cli.JmxTransConfiguration;
 import com.googlecode.jmxtrans.exceptions.LifecycleException;
+import com.googlecode.jmxtrans.executors.ExecutorRepository;
 import com.googlecode.jmxtrans.guice.JmxTransModule;
 import com.googlecode.jmxtrans.jobs.ServerJob;
 import com.googlecode.jmxtrans.model.JmxProcess;
@@ -56,8 +57,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.management.MBeanServer;
+import javax.management.MalformedObjectNameException;
 import java.io.File;
 import java.lang.management.ManagementFactory;
 import java.text.ParseException;
@@ -99,9 +102,13 @@ public class JmxTransformer implements WatchedCallback {
 	private Thread shutdownHook = new ShutdownHook();
 
 	private volatile boolean isRunning = false;
-	@Nonnull private final ThreadPoolExecutor queryProcessorExecutor;
-	@Nonnull private final ThreadPoolExecutor resultProcessorExecutor;
+	@Nonnull private ExecutorRepository queryExecutorRepository;
+	@Nonnull private ExecutorRepository resultExecutorRepository;
 	@Nonnull private final ThreadLocalRandom random = ThreadLocalRandom.current();
+	@Nonnull private final MBeanServer platformMBeanServer;
+	@Nullable private ManagedJmxTransformerProcess jmxTransformerProcessMBean;
+	@Nullable private ImmutableList<ManagedThreadPoolExecutor> queryExecutorMBeans;
+	@Nullable private ImmutableList<ManagedThreadPoolExecutor> resultExecutorMBeans;
 
 	@Inject
 	public JmxTransformer(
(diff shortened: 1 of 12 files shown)


gehel:
I need some time to read through that one... but first question: do we really want to keep the single thread pool? I think it is broken enough that we should remove it completely.

kipwoker:
I think we need to keep single strategy for old users.
For example, they can have queryProcessorExecutorPoolSize = 1000 and about 10 servers so they will have 10000 threads instead of 1000.
It's just a soft solution because I have no usage statistics. If you believe that this is unnecessary I can reduce that. But I think better to live with two strategies for a while and after 2 months, for example, drop common strategy.

Also I suggest to write it in wiki with explanation.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/executors/SeparateExecutorRepository.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.annotation.Nonnull;
+import javax.management.MalformedObjectNameException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Hashtable;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class SeparateExecutorRepository implements ExecutorRepository {
+	@Nonnull private final Hashtable<Server, ThreadPoolExecutor> repository;
```
gehel: I'd use a ConcurrentHashMap instead of a Hashtable. Slightly more efficient.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
```diff
@@ -317,14 +318,24 @@ public void executeStandalone(JmxProcess process) throws Exception {
...
 		this.processServersIntoJobs();
 	}
 
+	private void initializeExecutors() throws MalformedObjectNameException {
+		this.initializeExecutors(queryExecutorRepository);
+		this.initializeExecutors(resultExecutorRepository);
+	}
+
+	private void initializeExecutors(ExecutorRepository executorRepository) throws MalformedObjectNameException {
+		for (Server server : this.masterServersList) {
```
gehel: I'm wondering if it wouldn't make sense to make the ExecutorRepository lazy initialized (on a call to `getExecutor()`). This would allow to not have initialization logic outside of the Guava modules.

gehel:
I added a few comments inline, just to prove that I actually read the code :) Feel free to ignore them if you disagree. This PR looks good to me!

And thanks for planning to write doc! Doc is an area where we are clearly not as good as we should!

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
kipwoker: I thought JmxTransformer class has initialization responsobility, isn't it?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/executors/SeparateExecutorRepository.java
kipwoker: Okay. Fixed. :)

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
gehel: It does have some initialization responsibilities left, but I'm trying to slowly remove what's left when it make sense. I'm not sure it is worth it in this case...

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/executors/SeparateExecutorRepository.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.management.MalformedObjectNameException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class SeparateExecutorRepository implements ExecutorRepository {
+	@Nonnull private final ConcurrentHashMap<Server, ThreadPoolExecutor> repository;
+	@Nonnull private final ExecutorFactory executorFactory;
+	@Nonnull private final ArrayList<ManagedThreadPoolExecutor> mBeans;
```
gehel: mBeans should probably be thread safe. We are good accidentally, since initialization and shutdown are done in the same thread, but that's accidentally (I suspect the failing case is an edge case we'll never see, but better be safe than sorry). This means either making it a concurrent list (CopyOnWriteArrayList?) or immutable and move the initialization outside of the class. Or synchronize all access to mBeans...

Side note: the variable should be declared as `List`.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/executors/SeparateExecutorRepository.java
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	@Override
+	public ThreadPoolExecutor getExecutor(Server server) {
+		return repository.get(server);
+	}
+
+	@Override
+	public Collection<ManagedThreadPoolExecutor> getMBeans() {
+		return mBeans;
```
gehel: We don't want callers to modify that collection, so it would make sense to wrap it in a `Collections.unmodifiableList()`

gehel:
Thanks for the changes! Sorry, but I have to additional minor comments.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
kipwoker: I've tried to implement it in JmxModule, but I need server list. And server list initialization occurs over here.
I thought about extracting ServerRepository and inject it in ExecutorRepository but it seems more complicated to understand.

gehel: Understood. Not worth the effort. Let's move on!  :+1:

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/executors/SeparateExecutorRepository.java
kipwoker: You probably right. I'll make some changes.

gehel:
Looks all good! I'll just wait for the build to complete before merging. Thanks for the help!

kipwoker:
Thank you for review. :)

PR #632, opened 2017-12-20
Tabs vs spaces cleanup

jrglee:
Nothing fancy, just some cleanup while reading through the script. I am planning to add support for influxdb next.


gehel:
Thanks for the cleanup!

PR #641, opened 2018-02-07
Add new ZabbixWriterFactory for outputting metrics to Zabbix

jrbeilke:
Put together a new ZabbixWriterFactory that can send metrics to a Zabbix Server/Proxy using the Zabbix Sender protocol - https://www.zabbix.org/wiki/Docs/protocols#Zabbix_protocols

Here's an example config:
```
{
  "servers" : [ {
    "host" : "localhost",
    "alias" : "tomcat.dev",
    "port" : "1099",
    "queries" : [ {
      "obj" : "java.lang:type=Memory",
      "attr" : [ "HeapMemoryUsage", "NonHeapMemoryUsage" ],
      "outputWriters" : [ {
        "@class" : "com.googlecode.jmxtrans.model.output.ZabbixWriterFactory",
        "port" : 10051,
        "host" : "zabbix.dev"
      } ]
    } ]
  } ]
}
```

And on the Zabbix Server side a user will need to create trapper items matching the keys from jmxtrans (ZabbixWriter prefixes everything with 'jmxtrans.' to avoid collisions in Zabbix):
zabbix-jmxtrans-trapperitem

This output will work with the following Zabbix versions:
- Zabbix 2.0.8+
- Zabbix 2.1.7+
- Zabbix 2.2+
- Zabbix 3+

### DIFF
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/ZabbixWriter.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/ZabbixWriter.java
@@ -89,12 +89,12 @@ public class ZabbixWriter implements WriterBasedOutputWriter {
 				g.writeStringField("host", server.getHost());
 				g.writeStringField("key", key);
 				g.writeStringField("value", value.toString());
-				g.writeNumberField("clock", result.getEpoch());
+				g.writeNumberField("clock", result.getEpoch() / 1000);
 				g.writeEndObject();
 			}
 
 			g.writeEndArray();
-			g.writeNumberField("clock", System.currentTimeMillis());
+			g.writeNumberField("clock", System.currentTimeMillis() / 1000);
 			g.writeEndObject();
 			g.flush();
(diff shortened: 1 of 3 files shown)


jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/ZabbixWriter.java
```diff
@@ -0,0 +1,99 @@
...
+		) {
+			g.useDefaultPrettyPrinter();
+			g.writeStartObject();
+			g.writeStringField("request", "sender data");
+			g.writeArrayFieldStart("data");
+
+			for (Result result : results) {
+				log.debug("Query result: {}", result);
+
+				String zabbixHostName = server.getAlias() != null ? server.getAlias() : server.getHost();
```
gehel: This could be replaced by `Server.getLabel()`.

gehel:
Minor comment inline, but this looks good! Thanks!

gehel:
All looks good! Thanks a lot!

PR #645, opened 2018-02-22
Replace Log4J1 by Logback

gquintana:


gehel:
Great to see some work on getting rid of log4j! The tricky part will probably be to ensure that the current log4j based output writers are still working... and don't mess up the dependencies...

Have fun!

gehel:
I'm far from worried!

One solution to "fix" the log4j output writers, would be to delete them. We should be able to get the same functionalities with the [Slf4JOutputWriter](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/Slf4JOutputWriter.java) and the appropriate logback configuration. This would mostly need documentation.

We could also remove the log4j output writers from the main package and either let people who need them bundle them manually. Or provide an alternative package which still contains them.

In any case, those log4j output writers should die!

gquintana:
At this point, I completely removed the Log4J1 dependency and build is mostly successful.

However:
- there is one IT test I need to look at. I didn't change anything in the UdpOutputWriter. 
- there is LogBack dependency mismatch  warning during JACo prepare phase. I'll check whether upgrading Jacoco helps.
- I removed one Output writer because it got merged into DailyKeyOutputWriter
- The Slf4Jwriter in the LogBack module is really close to the one in Core module.
- The LogBack based outputwriters in the LogBack module start many things which are never stopped. These outputwriters have no unit test either.
- Once packaged as a fat jar (including LogBack dependencies), SLF4J doesn't find the LogBack implementation. The jar shading probably breaks the service provider config or something like that. Can I leave split jars in the tarbal?

gehel:
> there is one IT test I need to look at. I didn't change anything in the UdpOutputWriter.

Looks like a transient failure. I restarted the build. We might want to have a look at making that test more robust, but it does not look related to this PR.

> The Slf4Jwriter in the LogBack module is really close to the one in Core module.

I'm not sure I know which writers you are talking about. Could you add a link?

> Once packaged as a fat jar (including LogBack dependencies), SLF4J doesn't find the LogBack implementation. The jar shading probably breaks the service provider config or something like that. Can I leave split jars in the tarbal?

The shading does not rename anything, so I suspect the issue is related to duplicated classes. The [duplicate-finder-maven-plugin](https://github.com/basepom/duplicate-finder-maven-plugin) can help in this matter (and we should probably introduce it anyway, duplicated classes are always a mess...).

gquintana:
> I'm not sure I know which writers you are talking about. 
https://github.com/gquintana/jmxtrans/blob/pr-logback/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/Slf4JOutputWriter.java
https://github.com/gquintana/jmxtrans/blob/pr-logback/jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/LogbackWriter.java
The latter is poorly named as it has no dependency on Logback

> I suspect the issue is related to duplicated classes. 
It could also be a duplicated META-INF config (for Java 7+ service loader or the like). Understanding how the SLF4J plugin mechanisme works is the key.

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/LogbackWriter.java
```diff
@@ -88,7 +90,7 @@ public void internalWrite(Server server, final Query query, ImmutableList<Result
 
 				MDC.put("server", alias);
 				MDC.put("metric", KeyUtils.getKeyString(server, query, result, typeNames, null));
-				MDC.put("value", result.getValue());
+				MDC.put("value", result.getValue().toString());
```
gehel: All those MDC entries should be cleaned up after the call to log.info()

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/NagiosWriter.java
```diff
@@ -55,7 +57,7 @@
  */
 public class NagiosWriter extends BaseOutputWriter {
```
gehel: Why oh why does the NagiosWriter depends on log4j to do its job... We should really kill that dependency (and not replace it with a dependency on logback / slf4j). (but that might well be outside the scope of this PR).

gquintana: The NagiosWriter doesn't depend on Lo4J anymore.

This jmxtrans-output-module depends on Logback internals to create a RollingFileAppender per metric.

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/LogbackWriter.java
gquintana: I fixed it, but the bug was already in Log4J1 edition.
IMHO we should ditch the outputwriter in favor of jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/Slf4JOutputWriter.java

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/LogbackWriter.java
```diff
@@ -98,6 +98,7 @@ public void internalWrite(Server server, final Query query, ImmutableList<Result
 				MDC.put("valuePath", Joiner.on('.').join(result.getValuePath()));
 				MDC.put("Epoch", String.valueOf(result.getEpoch()));
 				log.info("");
+				MDC.clear();
```
gehel: Clearing the whole MDC is probably not the right thing to do. What if it already has some state, pushed by another piece of code. `MDC.put()` returns a `Closeable` that can be used to clear the specific values.

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/LogbackWriter.java
gehel: Yep, I'm all for ditching it!

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/NagiosWriter.java
gehel: I meant "Why oh why does the NagiosWriter depends on the internals of any logging framework". logback / log4j (or other), I don't really care. (But don't spend too much time on that cleanup, it probably belongs to another PR...)

gquintana: The NagiosWriter is vary close to KeyOutWriter, it uses a FileAppender. The only difference is log format.

gquintana:
I don't understand the problem with TravisCI

gehel:
> I don't understand precisely the problem when building jmxtrans-docker-test. 

What we do around docker at this point is a bit messy. We do have multiple builds (including one on docker hub), the Dockerfile downloads a specific version of jmxtrans (which might not be the latest). This part needs love!

> I think it's not a good idea to redirect stdout to the same file used by Logback, because it breaks log file format, it may cause to things to appear twice in the log file, it may block the Logback thread or the thread printing to stdout... Moreover this environment var LOG_FILE should be renamed STDOUT_FILE or something like this to avoid confusion.
> 
> I propose:
> - To remove the `export LOG_FILE="/var/log/jmxtrans/jmxtrans.log"`
> - Add an option in Slf4jWriter to log the result.toString() exactly like the StdoutWriter
> - Use this Slf4jWriter in jmxtrans-docker-test

All sounds reasonable.

> Concerning the failing build on TravisCI, I can not reproduce it locally on Ubuntu. How can I Know which Docker image is failing?

No idea. I did not do the setup of that part of the build (and never took the time to understand it properly). A `git blame` might give you a pointer to someone who knows better.

> Is it possible to build RPMs on Ubuntu?

Yes, the RPM are generated with the [rpm-maven-plugin](http://www.mojohaus.org/rpm-maven-plugin/), which is a pure java implementation. You should even be able to generate RPMs from Windows or MacOS (I have not tested). This was part of the reasoning for having packaging driven by Maven, instead of a more standard packaging process...

> Why is there a 120 minutes wait in Travis config?

This is actually a 120 timeout on the build, to override the standard travis timeout. The jmxtrans build takes way too much time, and we should really invest in making it faster (mostly optimizing tests, maybe removing pitest or tuning it).

gquintana:
Removing the opensuse13.2 image fixes the build on Travis build, but I don't know why. There must be something related to this particular image. Where can I find the Dockerfiles of jmxtrans/testing images?

> Yes, the RPM are generated with the rpm-maven-plugin, which is a pure java implementation.
I suspect there is some dependency to the rpm build tooling somewhere, because enabling the rpm profile make the build fail on my Ubuntu box.
```
[ERROR] Failed to execute goal org.codehaus.mojo:rpm-maven-plugin:2.1.5:attached-rpm (attach-rpm) on project jmxtrans: Unable to query for default vendor from RPM: Error while executing process. Cannot run program "rpm": error=2, Aucun fichier ou dossier de ce type -> [Help 1]
```

gehel:
> Removing the opensuse13.2 image fixes the build on Travis build, but I don't know why. There must be something related to this particular image. Where can I find the Dockerfiles of jmxtrans/testing images?

It looks like the Dockerfile is abstracted by the [docker-maven-plugin](https://github.com/fabric8io/docker-maven-plugin). And the installation is done at runtime with the [${rpm.run.cmd}](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-docker-test/pom.xml#L49). The jmxtrans-docker-test module is intended to do testing of the packaging, not to actually create a workable docker image.

The docker image available on [docker hub](https://hub.docker.com/r/jmxtrans/jmxtrans/) is built from https://github.com/jmxtrans/jmxtrans/tree/master/jmxtrans/docker (and has its own set of issues).

> I suspect there is some dependency to the rpm build tooling somewhere, because enabling the rpm profile make the build fail on my Ubuntu box.
> ```
> [ERROR] Failed to execute goal org.codehaus.mojo:rpm-maven-plugin:2.1.5:attached-rpm (attach-rpm) on project jmxtrans: Unable to query for default vendor from RPM: Error while executing process. Cannot run program "rpm": error=2, Aucun fichier ou dossier de ce type -> [Help 1]
> ```

My mistake... I probably installed the required dependencies a long time ago. It's the jdeb-maven-plugin which is pure Java...

Unrelated: I merged a few changes that will probably make rebasing this PR harder than it should have been. Apologies for that. I now realized that I should probably have waited a bit...

gquintana:
> It looks like the Dockerfile is abstracted by the docker-maven-plugin. And the installation is done at runtime with the ${rpm.run.cmd}. The jmxtrans-docker-test module is intended to do testing of the packaging, not to actually create a workable docker image.
>
> The docker image available on docker hub is built from https://github.com/jmxtrans/jmxtrans/tree/master/jmxtrans/docker (and has its own set of issues).

I'm not sure there is a relation between the image built in https://github.com/jmxtrans/jmxtrans/tree/master/jmxtrans/docker and the images service as base in https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-docker-test/pom.xml#L122

Both Dockerfile and docker-maven-plugin do the same, take a base image containing Java install the JMXTrans archive in it. The Dockerfile uses a standard openjdk:8 image, but the  docker-maven-plugin uses some custom base images like jmxtrans/testing:opensuse13.2-0.

Note for later: we should use docker-maven-plugin in jmxtrans module to build an official JMXTrans image

gehel:
@gquintana I'm following this PR, but not too closely... Ping me if you need review / merge...

gquintana:
@gehel the Java/Maven part should be OK. I am struggling with the packaging part and Docker testing. So far Travis CI is winning 😥

gehel:
Ok, I'll try to take time to have a deeper look. The docker integration tests are fairly recent and might have problems of their own. I have not looked at them too closely so far, so I'm probably not be of much help. I'll see if I can find something anyway...

gquintana:
Hooray build successful! @gehel  I'll reenable the docker test I've commented and I'm done

jmxtrans-core/pom.xml
```diff
@@ -134,6 +134,11 @@
 			<scope>system</scope>
 			<systemPath>${toolsjar}</systemPath>
 		</dependency>
+		<dependency><!-- quartz depends on Log4J -->
```
gehel: Do we really need this dependency? Wouldn't having it in the jmxtrans module be sufficient? Or maybe as a test dependency if we want to log from log4j during test. Since jmxtrans-core isn't defining a runtime artifact (it has to be packaged in the jmxtrans module), it seems to me that defining the logging implementation should be done somewhere else.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/Slf4JOutputWriterTest.java
```diff
@@ -31,21 +30,20 @@
...
 import static com.googlecode.jmxtrans.model.ResultFixtures.singleNumericResult;
+import static com.googlecode.jmxtrans.model.ResultFixtures.singleResult;
+import static com.googlecode.jmxtrans.model.ResultFixtures.stringResult;
 import static com.googlecode.jmxtrans.model.ServerFixtures.dummyServer;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
 
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(org.mockito.junit.MockitoJUnitRunner.class)
```
gehel: Why not importing this runner? I suspect that this is just your IDE trying to be smart and failing... (minor enough, feel free to ignore)

gehel:
That looks really good! Thanks for taking the time to cleanup all this!

I added a few comments inline, but feel free to ignore and merge on your own.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/Slf4JOutputWriterTest.java
gquintana: That's it. The MockitoJUnitRunner has moved in an another package, IntelliJ warns with deprecated

jmxtrans-core/pom.xml
gquintana: It's a not a logging implementation, it's a bridge:
```
quartz -> log4j-over-slf4j -> slf4j -> logback
```

gehel: Yep, wrong naming on my side. Still, since we are not using log4j in this module, but log4j comes from transitive dependencies, it looks to me like a packaging concern more suited for the jmxtrans module. But I don't have a strong opinion on this. Feel free to merge.

gehel:
I just finished the release... so sorry for the merge conflict. We now have a release just prior to this big refactoring, so we can merge it!

jmxtrans-core/pom.xml
gquintana: I see your point, I'll check if I can remove it.

jmxtrans-output/jmxtrans-output-core/src/test/java/com/googlecode/jmxtrans/model/output/Slf4JOutputWriterTest.java
gquintana: Fixed

jmxtrans-core/pom.xml
gquintana: @gehel You were right log4j-over-slf4j was useless because quart uses slf4j api (and log4j at runtime). It's just needed by Kafka and Zookeeper servers in integration tests.

gquintana:
I am not sure about commit 959c390 , I can revert it if you want

gquintana:
I discovered a small problem with docker tests. They print the log file content every second. When the CI is slow, it takes more time for the log to appear. As a result the CI log are bigger, as Travis kills the job when the log file is too big, in fact it doesn't wait enough for the log to appear.

I increased the polling to period to 5s, this reduces the verbosity and solves the problem. But it will make docker testing longer when the CI is fast. I'm not proud of this solution.

Another solution may be to not print the log file content (quick and less verbose). But it would make debugging harder.

gquintana:
@gehel can you give a last check before I merge, in particular the change in verify.sh. Thanks.

gehel:
\o/

Issue #644, opened 2018-02-20
Support for JMX notifications (events)


dgloeckner:
[JMX notifications](https://docs.oracle.com/javase/tutorial/jmx/notifs/index.html):

> The JMX API defines a mechanism to enable MBeans to generate notifications, for example, to signal a state change, a detected event, or a problem.

jmxtrans could be extended to support applications sending notifications / events via JMX.

Comment from Guillaume Lederrey:
> At the moment, jmxtrans does not have support for JMX events. That would be a great feature! It would require significant work, a new DSL to define events, but nothing impossible. Feel free to open an issue, and to submit a patch!

dgloeckner:
Let's see if I find time to support the project.

dgloeckner:
I've checked out how a JMX client can subscribe to JMX notifications.

The subscriber needs to
- know the name of the JMX bean wich he wants to subscribe to
- define a notification filter which decides if a particular Notification instance is passed to the subscriber
- implement NotificationListener interface

dgloeckner:
The listener interface contains only one method header:
`public void handleNotification(Notification notification, Object handback);`

AttributeChangeNotification is a sub class of Notification and it
> Provides definitions of the attribute change notifications sent by MBeans

jmxtrans could provide support for receiving AttributeChangeNotifications. A more generic approach might allow registering some kind of translator which is able to translate custom Notifications.

dgloeckner:
Short example for a subscriber which dumps all AttributeChangeNotifications to stdout.

```
package com.example.notifications;

import javax.management.*;
import javax.management.remote.JMXConnector;
import javax.management.remote.JMXConnectorFactory;
import javax.management.remote.JMXServiceURL;

public class Subscriber implements NotificationListener {

    private final String notificationType;

    public Subscriber(String notificationType) {
        this.notificationType = notificationType;
    }

    public void handleNotification(Notification notification, Object handback) {
        System.out.println(String.format("Listener %s: %s", notificationType, notification));
        if (notification instanceof AttributeChangeNotification) {
            AttributeChangeNotification attributeChangeNotification = (AttributeChangeNotification) notification;
            System.out.println(String.format("%s changed from %s to %s.",
                    attributeChangeNotification.getAttributeName(),
                    attributeChangeNotification.getOldValue(),
                    attributeChangeNotification.getNewValue()));
        }
    }

    public NotificationFilter getNotificationFilter() {
[...]

PR #646, opened 2018-03-01
Support for JMX notifications (events) #644

dgloeckner:
I've started to implement support for JMX notifications [as discussed in the issue](https://github.com/jmxtrans/jmxtrans/issues/644).

It's working end-to-end as can be seen in JmxTransformerIT integration test.

For now the only kind of Notification which is supported is AttributeChangeNotification. Applications could of course implement their own Notifications but jmxtrans would need to have related classes on its class path. It's probably not the common case. 

My use case is the following:
* Publish events (e.g. completed business processes with execution time and meta data).
* Stick to AttributeChangeNotifications, Java collections and primitives to model events.

Example publisher:
```Java
public class Hello
        extends NotificationBroadcasterSupport
        implements HelloMBean {

       ....
        Map<String, Object> vals = new HashMap<>();
        vals.put("cacheSize", this.cacheSize);
        vals.put("someId", "12345");
        vals.put("someVersion", "1.2.3");

        Notification n = new AttributeChangeNotification(this,
                sequenceNumber++, System.currentTimeMillis(),
                "Event occurred", "CustomEvent",   "Map",
                null, vals);

        sendNotification(n);
```

Example jmxtrans configuration:
```
          {
          "obj": "Examples:type=JMX,name=MXBean",
          "resultAlias": "myEvent",
          "queryType" : "NOTIFICATIONS",
          "attr": ["CacheSize"],
          "outputWriters": [
            {
              "@class": "com.googlecode.jmxtrans.model.output.InfluxDbWriterFactory",
              "url": "http://127.0.0.1:8086/",
              "username": "admin",
              "password": "admin",
              "database": "appstats",
              "tags": {
                "application": "my-app2"
              }
            }
          ]
        }
```
The query will subscribe to nofitications of the CacheSize attribute which are published by the configured MBean. queryType is set to NOTIFICATIONS to tell jmxtrans that we want to use events here and not polling. 

Example data from InfluxDB:
```
> select * from myEvent where time > now() - 1m
name: myEvent
time                 CacheSize CustomEvent_cacheSize CustomEvent_someId CustomEvent_someVersion _jmx_port application attributeName className hostname  objDomain typeName
----                 --------- --------------------- ------------------ ----------------------- --------- ----------- ------------- --------- --------  --------- --------
1522362599612000000            22                    12345              1.2.3                   10000     my-app2     CustomEvent   Hello     localhost Examples  type=JMX,name=MXBean
1522362601803000000            1                     12345              1.2.3                   10000     my-app2     CustomEvent   Hello     localhost Examples  type=JMX,name=MXBean
1522362612651000000            22                    12345              1.2.3                   10000     my-app2     CustomEvent   Hello     localhost Examples  type=JMX,name=MXBean
1522362630900000000            11                    12345              1.2.3                   10000     my-app2     CustomEvent   Hello     localhost Examples  type=JMX,name=MXBean
```

We can of course discuss the overall design and limitations. I'm keen to hear feedback.


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
```diff
@@ -30,31 +31,81 @@
...
+import java.util.ArrayList;
+import java.util.IdentityHashMap;
+import java.util.List;
 
 @ToString
 @ThreadSafe
 public class JMXConnection implements Closeable {
-	@Nullable private final JMXConnector connector;
-	@Nonnull @Getter private final MBeanServerConnection mBeanServerConnection;
+	@Nullable
```
gehel: It would be nice to not reformat code that you don't touch. It makes reviewing it much easier...

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
```diff
@@ -30,31 +31,81 @@
...
+	private static NotificationFilter getNotificationFilter(final ImmutableList<String> attributes) {
+		return new NotificationFilter() {
+			@Override
+			public boolean isNotificationEnabled(Notification notification) {
+				if (notification instanceof AttributeChangeNotification) {
+					// Check if subscribed attribute
+					AttributeChangeNotification changeNotification = (AttributeChangeNotification) notification;
+					return attributes.contains(changeNotification.getAttributeName());
+				}
+				return false;
```
gehel: Minor style issue: I prefer to have preconditions with early return instead of nested blocks. In this case, I think it would be nicer to have:

```
if (! notification instanceof AttributeChangeNotification) return false;
[...]
```

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
```diff
@@ -30,31 +31,81 @@
...
-	@Nonnull @Getter private final MBeanServerConnection mBeanServerConnection;
+	@Nullable
+	private final JMXConnector connector;
+	@Nonnull
+	@Getter
+	private final MBeanServerConnection mBeanServerConnection;
 	private boolean markedAsDestroyed;
 	private static final Logger logger = LoggerFactory.getLogger(JMXConnection.class);
 
+	private final IdentityHashMap<Object, QueryNotificationListener> notificationListeners = new IdentityHashMap<>();
```
gehel: Note that JMXConnections are pooled and should be considered ephemeral. Keeping a list of notificationListeneners as an instance member will probably cause issues when connections are recycled by the pool.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Query.java
```diff
@@ -79,10 +83,16 @@
...
+	public enum QueryType {
+		POLL,
+		NOTIFICATIONS;
+	}
+
 	private static final Logger logger = LoggerFactory.getLogger(Query.class);
 
 	/** The JMX object representation: java.lang:type=Memory */
 	@Nonnull @Getter private final ObjectName objectName;
+	@Getter private final QueryType queryType;
```
gehel: queryType should be @Nonnull

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Query.java
```diff
@@ -241,6 +256,41 @@ public String makeTypeNameValueString(List<String> typeNames, String typeNameStr
 		return ImmutableList.of();
 	}
 
+	public synchronized boolean isNotificationListenerRegistered(JMXConnection jmxConnection) {
```
gehel: Synchronization is probably not necessary here. The Query itself is immutable, so you're not protecting any state that is local to that class.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -278,10 +278,21 @@ private Server(
 		try {
 			jmxConnection = pool.borrowObject(this);
 			ImmutableList.Builder<Result> results = ImmutableList.builder();
-			MBeanServerConnection connection = jmxConnection.getMBeanServerConnection();
 
-			for (ObjectName queryName : query.queryNames(connection)) {
-				results.addAll(query.fetchResults(connection, queryName));
+			if(query.getQueryType() == Query.QueryType.POLL) {
```
gehel: An Enum looks like a good candidate for a `switch` statement instead of `if`s.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -278,10 +278,21 @@ private Server(
...
+				// Fetch results from mbeans
+				MBeanServerConnection connection = jmxConnection.getMBeanServerConnection();
+
+				for (ObjectName queryName : query.queryNames(connection)) {
+					results.addAll(query.fetchResults(connection, queryName));
+				}
+			} else {
+				// Subscribe to notifications
+				if(!query.isNotificationListenerRegistered(jmxConnection)) {
```
gehel: A notification should probably not be executed, but registered at startup. The processing of results should be done when the notification is received, not polled. Basically, the notification listener should directly send results to the ResultProcessor...

jmxtrans-test-utils/src/main/java/com/googlecode/jmxtrans/test/OutputCapture.java
```diff
@@ -52,6 +52,9 @@ protected void before() throws Throwable {
 	protected void after() {
 		System.setOut(originalOut);
 		System.setErr(originalErr);
+		// I guess it's useful to print this after completion
+		System.out.println("What was captured:");
```
gehel: I would disagree... It is the job of the client of OutputCapture to print stuff if it make sense. In most cases, it probably does not. And even if it does, it should be sent to a logger.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/QueryNotificationListener.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Collects received notifications for a query.
+ * Notifications should be fetched via {@link #dumpNotifiations(List)}.
+ * TODO: if no one is calling dumpNotifiations() but we continue to receive
+ * notifications, jmxtrans would eventually go into OOM
+ */
+class QueryNotificationListener implements NotificationListener {
+
+	private final List<Notification> notifications = new ArrayList<>();
```
gehel: You probably want to use a concurrent queue instead of an ArrayList here. You could then get rid of all the synchronization of this class.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
dgloeckner: Sorry was a mistake and yes makes reviewing a lot easier ;) I'll double check next time.

jmxtrans-test-utils/src/main/java/com/googlecode/jmxtrans/test/OutputCapture.java
dgloeckner: I just found that debugging the integration test was painful. After adding the above output, I was able to see that my test needed to be fixed.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
dgloeckner: The design with polling was the simplest starting point I could think of. Just for my understanding, why do you think polling is not a good idea in this case?

Pushing (converted) notifications to registered ResultProcessors could also be implemented. 

I need to understand
- where I could register the notification listeners at startup
- how to retrieve the output writers for a given query

Are output writers thread safe?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
dgloeckner: Sure... final cleanup pending :)

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Query.java
dgloeckner: Yes.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
dgloeckner: In the design with polling (see the other comment) I needed to map received notifications to the query (List<Notification> getNotifications(Object query)). We would not need such a mapping if we implement a push-based approach as you suggested. Let's see how it goes :)

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
dgloeckner: Besides the style issue, do you think that limiting the support to AttributeChangeNotifications is OK for now?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
gehel: No problem, that's really a minor issue...

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
gehel: I have a feeling that supporting other kind of notification could have value. But starting with just AttributeChange and building up from there is perfectly fine. We can always improve later if there is a need.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
gehel: I think that the real issue is that JMX notifications rely on a long lived JMX connection. Which to some extend goes against the model we have. And in all cases, JMX connection can break for a number of reason (transient network issues seems the most obvious). So we need to account for that in some way. Not an easy task.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
gehel: Polling an intermediate list of notification requires to keep that state for some time. As you noted in your comment, that state can grow unbounded and end in an OOM (so at least that queue should be bounded). Also, having a queue of notification will delay those notifications from being processed for a somewhat arbitrary reason. If in the end, we're functionally just doing polling of an attribute value, then why not do just that and use the current query mechanics.

jmxtrans-test-utils/src/main/java/com/googlecode/jmxtrans/test/OutputCapture.java
gehel: I think that in this case, you should probably add the output to the integration test itself, via a logger. Or use a debugger when debugging, step through the code and check the state...

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
dgloeckner: I agree to the points about polling. Whether the delay is acceptable depends on the use case I guess. For me having a few seconds delay is acceptable. I disagree that polling from a list of received notifications is equivalent to polling MBeans. In the former case, the publisher sends explicit events and we should receive all of them (if we stay connected). This is not guaranteed with polling of MBeans as we might miss intermediate attribute values.

One problem with the pushing design is that we need to re-register listeners when re-creating JMX connections. I guess MBeanServerConnectionFactory can't do this for us. On the other hand, it's the only one who knows when a connection is created. I was thinking that the initial listener registration could happen from JmxTransformer when we iterate over servers and queries. But how to handle broken and re-created connections is still a mistery to me. Would you have any idea?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/connections/JMXConnection.java
dgloeckner: Yes, we need to re-create listeners when re-creating connections. We would need to detect the liveness of JMX connections and deal with listener re-creation. Any hints appreciated :)

dgloeckner:
Hi,

I've just pushed a commit where I changed my design to use pushing instead of polling. 

Here's how it works:
- when a new JMX connection is created for a given server (JmxConnectionProvider) the notification listeners are registered. MBeanServerConnectionFactory calls  server.subscribeToNotifications() when a connection is created.
- NotificationProcessor receives the notifications, converts them to results and passes them to the configured output writers. On the calling thread atm.

Some questions popped up:
- query.queryNames() - what is this used for and do I use it correctly in Query.subscribeToNotifications()
- Threading / concurrency: see comment in NotificationProcessor

Keen to hear your feedback.

Cheers,
Daniel

gquintana:
See https://github.com/jmxtrans/jmxtrans/pull/648 for problem with AppVeyorCI

gquintana (commented):
Interesting

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Query.java
```diff
@@ -241,6 +276,48 @@ public String makeTypeNameValueString(List<String> typeNames, String typeNameStr
...
+			Object handback = notificationProcessorFactory.create(server, this, oi);
+			NotificationListener notificationListener = new NotificationListener() {
+				@Override
+				public void handleNotification(Notification notification, Object handback) {
+					NotificationProcessor handler = (NotificationProcessor) handback;
+					handler.handleNotification(notification);
+				}
+			};
+
+			jmxConnection.addNotificationListener(oi.getObjectName(),
```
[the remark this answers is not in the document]
dgloeckner: In case of a broken connection the monitored app would need to remove the (dead) listeners. I haven't checked the JMX server code yet but there must be some logic for that.

dgloeckner:
@gquintana @gehel I appreciate if you could have a look at my questions posted after the latest commit :)

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Query.java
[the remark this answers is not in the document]
gehel: There isn't any logic yet. It is probably possible to plug into the connection pool to process the removal of listener before a JMXConnection is killed.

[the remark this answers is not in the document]
dgloeckner: Which connection pool are you referring to here? From our pov the monitored app is a pure JAVA app which exposes metrics via JMX. About listener removal in jmxtrans, I'm taking care of it (I hope).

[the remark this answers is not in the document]
gehel: The JMX Connections are pooled. The pool is managed [by a Guice module](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/guice/JmxTransModule.java#L90-L91), and is used in [Server.execute()](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java#L276-L300). Having a pool for the JMX connection is mainly useful to ensure that connections are validated, recycled, and overall managed correctly. It make things a bit awkward for notifications. You can't assume that a there is a single JMX Connection to any application. You also can't assume that those connections are long lived (well, you should never assume they are, things can and will break, connections will be reset outside of your control).

gehel:
A few comments:

> when a new JMX connection is created for a given server (JmxConnectionProvider) the notification listeners are registered. MBeanServerConnectionFactory calls server.subscribeToNotifications() when a connection is created.

That is problematic as multiple JMX Connections can be opened to a single application (see inline comment about connection pool). We need to abstract the subscription even more. Also I think that the pool has a default min idle of 0, which means that the connections are created only on demand. If we only have notification queries, the connection will probably never be created and the notifications never registered. 

> NotificationProcessor receives the notifications, converts them to results and passes them to the configured output writers. On the calling thread atm.
Epoch (timestamp) is taken from the notification itself.

That looks fine.

> Some questions popped up:
> 
> query.queryNames() - what is this used for and do I use it correctly in Query.subscribeToNotifications()

Yes, you seem to use it correctly. A query can resolve to multiple mbeans. `queryNames()` will execute the query and return a list of the ObectNames of the mbeans matching the query.

> Threading / concurrency: see comment in NotificationProcessor
> objectInstance in JmxResultProcessor: am I passing in the correct object? 

I'm not sure what you are referring to. Could you add a link?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/NotificationProcessor.java
```diff
@@ -0,0 +1,29 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.googlecode.jmxtrans.model;
+
+import javax.management.Notification;
+
+public interface NotificationProcessor {
```
gehel: This interface has a single implementation, and does not provide any additional abstraction. Just merge it with the implementation class.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/NotificationProcessorFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.googlecode.jmxtrans.model;
+
+import javax.management.ObjectInstance;
+
+public interface NotificationProcessorFactory {
```
gehel: Same here, this interface does not add any value, please remove it.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/notifications/NotificationProcessorImpl.java
```diff
@@ -0,0 +1,102 @@
...
+				}
+			}
+		};
+		executor.submit(task);
+	}
+
+	private List<Result> transform(Notification notification) {
+		if (notification instanceof AttributeChangeNotification) {
+			AttributeChangeNotification changeNotification = (AttributeChangeNotification) notification;
+			List<Attribute> attributes = new ArrayList<>(1);
```
gehel: Use `Collections.singletonList()` instead of an explicitly sized ArrayList.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/JmxResultProcessor.java
```diff
@@ -46,13 +46,19 @@
 	private final String className;
 	private final String objDomain;
 	private final List<Attribute> attributes;
+	private final long epoch;
 
 	public JmxResultProcessor(Query query, ObjectInstance objectInstance, List<Attribute> attributes, String className, String objDomain) {
+		this(query, objectInstance, attributes, className, objDomain, System.currentTimeMillis());
```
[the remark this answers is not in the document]
gehel: Agreed! Also, don't use `System.currentTimeMillis()`, there is a `Clock` abstraction in jmxtrans, you should use it. (and we should replace it with `java.time.Clock` at some point).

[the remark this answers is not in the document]
gehel: Re-reading that class again, it is a mess (it was already before you touched it!). The time of query should be injected, not instantiated here.

gehel:
Despite all the comments, congratulation on the nice job! You are trying to address a complex problem, within a code base that could be much cleaner than it is! I hope we'll manage to make all that work!

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/NotificationProcessor.java
dgloeckner: Needed to avoid package cycle. You might suggest another design but there are similar examples in jmxtrans already (also with single implementation):

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/NotificationProcessorFactory.java
dgloeckner: Needed to avoid package cycle. You might suggest another design but there are similar examples in jmxtrans already (also with single implementation):

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Query.java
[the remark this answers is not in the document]
dgloeckner: > You can't assume that a there is a single JMX Connection to any application.
That's true, but can I assume that there's a single JMX connection per configured server (might be multiple "server configs" for the same application)? If yes, my design should work.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/NotificationProcessor.java
gehel: jmxtrans is full of examples of things you should never do :) I'll have a look to see if I can find a better solution.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Query.java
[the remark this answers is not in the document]
gehel: Nope, you can't even assume that there is a single connection per server. The [GenericKeyedObjectPool](https://commons.apache.org/proper/commons-pool/api-2.5.0/org/apache/commons/pool2/impl/GenericKeyedObjectPool.html) that we use works as a map of sub pools. This allows multiple threads to work in parallel to execute queries. Since JMXConnections are threadsafe and probably have very low contention, we could use a smarter way of managing them, getting away from the KeyedObjectPool and implementing connection validation / recycling. That might be a solution, but it requires a non trivial amount of work.

dgloeckner:
@gehel I've just pushed a major change. There's now an explicit object pool which handles a notification processor per server-query tuple. A notification processor has a private JMX connection so managing the subscriptions (also with varying nr of returned object names) is much easier. The notification listeners will be subscribed / re-subscribed at a regular interval.

dgloeckner:
I'm sure there are more reviews for "cosmetic" changes. I hope we get closer to the final design now and than the fine tuning can start.

dgloeckner:
Here's the gist of the re-design:
```Java
    public void subscribeToNotifications(final Server server) throws Exception {
		NotificationProcessor processor = null;
		ServerQuery serverQuery = new ServerQuery(server, this);
		try {
			processor = notificationProcessors.borrowObject(serverQuery);
			processor.subscribeToNotifications();
		} catch (Exception e) {
			if(processor != null) {
				notificationProcessors.invalidateObject(serverQuery, processor);
			}
			throw e;
		} finally {
			notificationProcessors.returnObject(serverQuery, processor);
		}
	}
```
Calling processor.subscribeToNotifications() takes care of adding new notification listeners and removing notification listeners for which the remote object name does not exist any more. The notification processor contains a JMXConnection. If anything goes wrong during subscribeToNotifications(), tell the object pool to invalidate the processor. The object pool is configured with max active = 1 so the periodic call to server. subscribeToNotifications() will always get the same notification processor for a given server - query tuple. 

What I like is that the state for notification listeners is isolated in one object, the NotificationProcessor. That object is responsible for bookkeeping. It features "let it crash" as we can simply throw away a notification processor which threw an exception (possibly due to a connection issue) and we'll try again from scratch without any extra code.

Calling processor.subscribeToNotifications() periodically also acts as a "heart beat" to ensure that the JMX connection is still alive so we can be sure that we're able to receive notifications. It also takes care of new object names appearing at run time which could match our query.

dgloeckner:
TODOs:
- have design approved ;)
- add unit tests and extend integration test further (e.g. MXBeans added / removed at runtime, simulate broken connections)
- fix all issues from code review once the design is approved
- document usage, design and limitations

dgloeckner:
@gehel would be cool if you could review the re-design, specifically handling of JMX connections. Maybe one JMX connection per server and query to receive notifications is too extreme. On the other hand, I wanted to keep state management simple which is a feature of the current design.

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -178,7 +178,7 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
 
 			HashMap<String, Object> filteredValues = newHashMap();
 			Object value = result.getValue();
-			if (isValidNumber(value)) {
+			if (isValidNumber(value) || value instanceof String) {
```
dgloeckner: It seems strange to only allow "valid numbers". InfluxDB for sure [allows to store Strings](https://docs.influxdata.com/influxdb/v1.5/write_protocols/line_protocol_reference/).

tjeppesen:
Was this merged into JMXTrans? Does the latest JMXTrans release support mbean notifications?

PR #648, opened 2018-03-03
Fix AppVeyorCI build

gquintana:
http://www.us.apache.org/ mirror doesn't archive Maven versions :'(


gehel:
I would suggest #650 as a slightly more standard option for the same result...

gquintana:
The Maven Wrapper is a brilliant solution to this problem. <3

gehel:
Brilliant might be a strong word, but yeah, it is nice...

PR #656, opened 2018-03-13
Going for openjdk-apline + automate last jmxtrans version

waldosCH:
The goal is to reduce the size of the image.

### DIFF
--- a/jmxtrans/docker/Dockerfile
+++ b/jmxtrans/docker/Dockerfile
@@ -64,7 +64,7 @@ RUN mkdir -p ${JMXTRANS_HOME}/conf
 COPY logback.xml ${JMXTRANS_HOME}/conf/logback.xml
 
 RUN mkdir -p /usr/share/jmxtrans/lib/ \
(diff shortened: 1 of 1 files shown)


jmxtrans/docker/Dockerfile
```diff
@@ -69,14 +55,16 @@ ENV CONTINUE_ON_ERROR false
...
+RUN addgroup jmxtrans \
+   && adduser jmxtrans -s /bin/bash -h /usr/share/jmxtrans -S -D -G jmxtrans
 
 WORKDIR ${JMXTRANS_HOME}
 RUN mkdir -p ${JMXTRANS_HOME}/conf
 
 COPY logback.xml ${JMXTRANS_HOME}/conf/logback.xml
 
 RUN mkdir -p /usr/share/jmxtrans/lib/ \
+    && JMXTRANS_VERSION=`curl -s http://central.maven.org/maven2/org/jmxtrans/jmxtrans/ | grep -v maven-metadata | grep '<a' | tail -n 1 | sed -r 's/.*href="([^\/]+).*/\1/'` \
```
gehel: Instead of parsing the HTML, which is meant for human consumption, not machine consumption, you should parse the maven-metadata.xml. The content of the <release/> tag should give you the latest release. I think that <latest/> could contain a snapshot (well, central does not host snapshots, so in this case, they are probably always equals).

Using [jq](https://stedolan.github.io/jq/) to parse xml is nicer, but it brings another dependency. I'm fine either way (with a slight preference for adding jq).

Issue #649, opened 2018-03-04
Build on Java 9


gquintana:
Identified impacts:
- Maven Cobertura Plugin doesn't support Java 9 yet as it has a dependency on tools.jar (using Jacoco may help)
- Maven Failsafe Plugin requires an upgrade to 2.20.1
- JHadeIT tests fails (don't know why):
```
[ERROR] Errors: 
[ERROR]   JHadesIT>JHadesBaseTest.classLoaderInfo:35 » NoClassDefFound sun/misc/Launcher
[ERROR]   JHadesIT>JHadesBaseTest.classpathInfo:41 » NoClassDefFound sun/misc/Launcher
[ERROR]   JHadesIT>JHadesBaseTest.multipleClassVersionsReport:46 » NoClassDefFound sun/m...
[ERROR]   JHadesIT>JHadesBaseTest.overlappingJarsReport:51 » NoClassDefFound sun/misc/La...
```
- com.sun:tool dependency doesn't exist anymore
```
		<dependency>
			<groupId>com.sun</groupId>
			<artifactId>tools</artifactId>
			<version>1.6.0</version>
			<scope>system</scope>
			<systemPath>${toolsjar}</systemPath>
		</dependency>
```
- Server class uses com.sun.tools.attach.VirtualMachine to connect to JVM using the PID
- Findbugs prints many warnings (upgrading to Spotbugs may help):
```
[INFO] --- findbugs-maven-plugin:3.0.5:findbugs (findbugs) @ jmxtrans-output-core ---
[INFO] Fork Value is true
     [java] WARNING: An illegal reflective access operation has occurred
     [java] WARNING: Illegal reflective access by org.dom4j.io.SAXContentHandler (file:/home/gquintana/local/maven3-repo/dom4j/dom4j/1.6.1/dom4j-1.6.1.jar) to method com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser$LocatorProxy.getEncoding()
     [java] WARNING: Please consider reporting this to the maintainers of org.dom4j.io.SAXContentHandler
     [java] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
     [java] WARNING: All illegal access operations will be denied in a future release
     [java] The following classes needed for analysis were missing:
     [java]   java.lang.Object
     [java]   java.lang.Enum
     [java]   java.util.concurrent.Callable
```

- Findbugs finds a new problem (this one is really easy to fix):
```
[...]

PR #658, opened 2018-03-13
[WiP] Build on Java 9

gquintana:
Replace Cobertura with Jacoco
Fix tools.jar dependency
Remove SignalInterceptor (seems unused and uses sun.*)


jmxtrans-core/pom.xml
```diff
@@ -179,5 +172,23 @@
...
+		<profile>
+			<id>toolsJar</id>
+			<activation>
+				<file>
+					<exists>${toolsjar}</exists>
+				</file>
+			</activation>
+			<dependencies>
+				<dependency>
+					<groupId>com.sun</groupId>
```
gehel: I'm not sure where this dependency came from and if we even need it anymore. Could we just drop it? Or could you add a comment on why it is needed (if you know)?

gquintana: We can't drop it see issue #649 where I not the problems I encounter.

gehel: Of course!

PR #659, opened 2018-03-14
Docker with alpine - missing libc6-compat 

waldosCH:


gehel:
The last commit looks good, but the first 2 have already been merged. Could rebase / cherrypick that last commit on top of master? Thanks!

gquintana:
We should use the fabric8io/docker-maven-plugin to build Docker images. It would avoid downloading the jmxtrans jar from Maven central repository. It could use the current (in development) jar instead. 
It would allow building several images (not everyone likes alpine) from the same jar and pushing them to a Docker registry.

gehel:
@gquintana yep, doing everything from maven would be much cleaner! And would probably ensure that we see errors in docker builds (I had a look after @waldosCH last PR, and the situation is not good. I just never look at those docker builds).

Now, someone just has to find time to put this in place :)

PR #660, opened 2018-03-15
Adding : typeNamesAsTags and segregateStringValues for InfluxDbWriter

yassineLajmi:
typeNamesAsTags : when this option is true, the writer will add typeNames as separate tags.
segregateStringValues : when the field value is of type non numerical, the Writer will add a anther field with _Str (to avoid the issue of influxData)

### DIFF
--- a/jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
+++ b/jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
@@ -32,6 +32,7 @@ import com.googlecode.jmxtrans.model.Result;
 import com.googlecode.jmxtrans.model.ResultAttribute;
 import com.googlecode.jmxtrans.model.Server;
 import com.googlecode.jmxtrans.model.naming.KeyUtils;
+import com.googlecode.jmxtrans.model.naming.typename.TypeNameValue;
 import org.influxdb.InfluxDB;
 import org.influxdb.InfluxDB.ConsistencyLevel;
 import org.influxdb.dto.BatchPoints;
@@ -78,6 +79,8 @@ public class InfluxDbWriter extends OutputWriterAdapter {
 	private final ImmutableSet<ResultAttribute> resultAttributesToWriteAsTags;
 
 	private final boolean createDatabase;
+	private final boolean typeNamesAsTags;
+	private final boolean segregateStringValues;
 	private final boolean reportJmxPortAsTag;
 
 	private final Predicate<Object> isNotNaN = new Predicate<Object>() {
@@ -96,7 +99,9 @@ public class InfluxDbWriter extends OutputWriterAdapter {
 			@Nonnull ImmutableSet<ResultAttribute> resultAttributesToWriteAsTags,
 			@Nonnull ImmutableList<String> typeNames,
 			boolean createDatabase,
-			boolean reportJmxPortAsTag) {
+			boolean reportJmxPortAsTag,
+			boolean typeNamesAsTags,
+			boolean segregateStringValues) {
 		this.typeNames = typeNames;
 		this.database = database;
 		this.writeConsistency = writeConsistency;
@@ -106,6 +111,8 @@ public class InfluxDbWriter extends OutputWriterAdapter {
 		this.resultAttributesToWriteAsTags = resultAttributesToWriteAsTags;
 		this.createDatabase = createDatabase;
 		this.reportJmxPortAsTag = reportJmxPortAsTag;
+		this.typeNamesAsTags = typeNamesAsTags;
+		this.segregateStringValues = segregateStringValues;
 	}
 
 	/**
@@ -178,9 +185,20 @@ public class InfluxDbWriter extends OutputWriterAdapter {
 
 			HashMap<String, Object> filteredValues = newHashMap();
 			Object value = result.getValue();
+
+			ImmutableList<String> typeNamesParam = this.typeNames;
+			// if typeNamesAsTag, we don't concat typeName in values.
+			if (typeNamesAsTags) {
+				typeNamesParam = ImmutableList.<String>of();
+			}
+
+			String key = KeyUtils.getPrefixedKeyString(query, result, typeNames);
 			if (isValidNumber(value)) {
-				String key = KeyUtils.getPrefixedKeyString(query, result, typeNames);
 				filteredValues.put(key, value);
+			}else {
+				if (segregateStringValues) {
+					filteredValues.put(key + "_str", value);
+				}
 			}
 
 			// send the point if filteredValues isn't empty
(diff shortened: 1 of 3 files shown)


jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -178,9 +185,20 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
...
+			if (typeNamesAsTags) {
+				typeNamesParam = ImmutableList.<String>of();
+			}
+
+			String key = KeyUtils.getPrefixedKeyString(query, result, typeNames);
 			if (isValidNumber(value)) {
-				String key = KeyUtils.getPrefixedKeyString(query, result, typeNames);
 				filteredValues.put(key, value);
+			}else {
+				if (segregateStringValues) {
```
gehel: If `value` is not a valid number, it does not mean it is a `String`. It can still be any kind of object. So there is at least some naming issue with `segregateStringValues`, possibly a conceptual issue as well.

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -178,9 +185,20 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
...
+					filteredValues.put(key + "_str", value);
```
gehel: Up to now, we were just filtering out non numbers (probably a wrong assumption that most backen only support numbers). If InfluxDB actually supports non number values, I'm not sure why we need that `_str` suffix, why not just push any value to influx?

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -209,6 +227,17 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
 			resultAttribute.addTo(resultTagMap, result);
 		}
 
+		if (typeNamesAsTags) {
+			log.debug("adding typeNames as tags");
+			Map<String, String> typeNameValueMap = TypeNameValue.extractMap(resultTagMap.get("typeName"));
+			log.debug("this is typeNameValueMap {} ==== {} ", typeNameValueMap, this.typeNames);
+			for (String typeToTag : this.typeNames) {
+				if (typeNameValueMap.containsKey(typeToTag)) {
```
gehel: Replace the `containsKey()` call with a get and check for nullity. This save one lookup in the map.

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
yassineLajmi: We have metrics that change from time to time, and influx does not like that, and if a metric is sent at the beginning in boolean, it will not be accepted anymore if it corrects itself and sends its numerical data.

putting everything in String is for simplicity, and can have a history of values sent in error (String and boolean)

yassineLajmi: is it Ok if I rename to segregateNoNumValues ?

gehel: For the boolean case, there is a [BooleanAsNumberValueTransformer](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/results/BooleanAsNumberValueTransformer.java) that might need to be added to the InfluxDB writer.

For the more generic case of segregating non numeric values, yes, your solution seems to make sense. Please rename the parameter as `segregateNonNumericValues` and add a parameter for the suffix. Also, a few lines of javadoc on the parameter explaining the use case would be welcomed!

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -209,6 +227,17 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
 			resultAttribute.addTo(resultTagMap, result);
 		}
 
+		if (typeNamesAsTags) {
+			log.debug("adding typeNames as tags");
+			Map<String, String> typeNameValueMap = TypeNameValue.extractMap(resultTagMap.get("typeName"));
+			log.debug("this is typeNameValueMap {} ==== {} ", typeNameValueMap, this.typeNames);
+			for (String typeToTag : this.typeNames) {
+				if (typeNameValueMap.get(typeToTag) != null) {
```
gehel: The point of using `get()` was to remove one lookup into the map, so you'll need to store that result and not do the same `get()` again on the next line :)

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
gquintana: I agree. Why do you add this "_str" suffix to the metric name? 
In the current version JMXTrans version non numeric values are not sent to InfluxDB, so the value type should not flicker.

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -209,6 +227,17 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
 			resultAttribute.addTo(resultTagMap, result);
 		}
 
+		if (typeNamesAsTags) {
+			log.debug("adding typeNames as tags");
```
gquintana: This debug logs seem useless

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
gehel: My understanding is that @yassineLajmi has mbeans that for the same attribute will sometime return a boolean and sometime an integer. The change might happen with a new versio of the application under monitoring, or just with a badly written mbean.

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
gehel: My understanding is that @yassineLajmi is interested in collecting those non numeric values for analysis. Since the backend allows collecting non numeric values (at least, that's my understanding), collecting those directly might make sense. I'm not thrilled by the use case, but I can see a need for it.

A better solution would probably to centralize the jmxtrans logs and treat this as a failure of metric collection...

nlaurent9816:
I think this PR can now be closed: the commits were updated and included in the PR #681.

PR #663, opened 2018-04-03
Fixed : Timestamp in milliseconds

nlaurent9816:
Graphite plaintext protocol uses a timestamp in secondes, but the writer writes this timestamp in ms. I just divided the value by 1000 to fit the protocol.

### DIFF
--- a/jmxtrans-output/jmxtrans-output-logback/pom.xml
+++ b/jmxtrans-output/jmxtrans-output-logback/pom.xml
@@ -69,6 +69,10 @@
 			<artifactId>jmxtrans-utils</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.jmxtrans</groupId>
+			<artifactId>jmxtrans-output-core</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
--- a/jmxtrans-output/jmxtrans-output-logback/pom.xml
+++ b/jmxtrans-output/jmxtrans-output-logback/pom.xml
@@ -66,11 +66,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.jmxtrans</groupId>
-			<artifactId>jmxtrans-utils</artifactId>
+			<artifactId>jmxtrans-output-core</artifactId>
 		</dependency>
 		<dependency>
 			<groupId>org.jmxtrans</groupId>
-			<artifactId>jmxtrans-output-core</artifactId>
+			<artifactId>jmxtrans-utils</artifactId>
 		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
--- a/jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/KeyOutWriter.java
+++ b/jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/KeyOutWriter.java
@@ -78,9 +78,8 @@ public class KeyOutWriter extends BaseOutputWriter {
 	protected static final String MAX_LOG_FILE_SIZE = "10MB";
 	protected static final String DEFAULT_LOG_PATTERN = "%msg%n";
 	protected static final String DEFAULT_DELIMITER = "\t";
-	protected LogWriter logwriter;
-
-
+	
+	private LogWriter logwriter;
 	private final String outputFile;
 	private final String maxLogFileSize;
 	private final int maxLogBackupFiles;
@@ -158,8 +157,8 @@ public class KeyOutWriter extends BaseOutputWriter {
 	 */
 	public static class LogWriter extends Writer{
 		
-		private Logger logger;
-		private String delimiter;
+		private final Logger logger;
+		private final String delimiter;
 		
 		public LogWriter(Logger logger, String delimiter) {
 			this.logger = logger;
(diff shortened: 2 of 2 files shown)


gehel:
If you are trying to send data to Graphite, you should probably use the [GraphiteWriter(2)](https://github.com/jmxtrans/jmxtrans/blob/master/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/GraphiteWriterFactory.java).

Since the KeyOutWriter claims to use the same format as Graphite, there is actually a bug here. To ensure the formats are the same, we should have the KeyOutWriter delegate to the GraphiteWriter instead of reimplementing the format.

@nlaurent9816 any chance you could modify this PR to reuse the GraphiteWriter2 instead of reimplementing the format?

Thanks!

nlaurent9816:
@gehel 
I have a specific use case where I need to write the Graphite metrics in a file before sending them to the Graphite server, that's why I use the KeyOutWriter.

I will try to reimplement this format using GraphiteWriter2 if you want. Do you prefer a new class KeyOutWriter2 or erase the actual KeyOutWriter ?

gehel:
Just modifying the implementation of the internalWrite() method of the current KeyOutWriter should be enough. Since it would not change its behaviour (except for fixing this bug) no need to create a new class.

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/KeyOutWriter.java
```diff
@@ -147,21 +147,52 @@ public void validateSetup(Server server, Query query) throws ValidationException
...
-		List<String> typeNames = getTypeNames();
-
-		for (Result result : results) {
-			if (isNumeric(result.getValue())) {
-
-				logger.info(KeyUtils.getKeyString(server, query, result, typeNames, null) + delimiter
-						+ result.getValue().toString() + delimiter + result.getEpoch());
-			}
-		}
+		LogWriter logwriter = new LogWriter(logger,delimiter);
```
gehel: This logwriter could be a field, no need to instantiate a new one for each message.

jmxtrans-output/jmxtrans-output-logback/pom.xml
```diff
@@ -68,6 +68,10 @@
 			<groupId>org.jmxtrans</groupId>
 			<artifactId>jmxtrans-utils</artifactId>
 		</dependency>
+		<dependency>
```
gehel: pom.xml should be sorted according to standard. You can use `./mvnw sortpom:sort` to sort it automatically.

nlaurent9816: Thanks for the reviews! I will correct all these mistakes.

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/KeyOutWriter.java
```diff
@@ -80,14 +78,15 @@
 	protected static final String MAX_LOG_FILE_SIZE = "10MB";
 	protected static final String DEFAULT_LOG_PATTERN = "%msg%n";
 	protected static final String DEFAULT_DELIMITER = "\t";
-	protected Logger logger;
+	protected LogWriter logwriter;
```
gehel: Why is this `protected` and not `private final` ? Any good reason?

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/KeyOutWriter.java
```diff
@@ -130,38 +130,70 @@ public KeyOutWriter(
...
+	
+	/**
+	 * Simple class which extends the Writer Class in order to use
+	 * our logger with the GraphiteWriter2 implementation.
+	 * @author a690062
+	 *
+	 */
+	public static class LogWriter extends Writer{
+		
+		private Logger logger;
```
gehel: looks like those 2 fields could be `final`

gehel:
Sorry, a few minor additional comments. But this looks almost ready to merge! Thanks a lot!

jmxtrans-output/jmxtrans-output-logback/src/main/java/com/googlecode/jmxtrans/model/output/KeyOutWriter.java
nlaurent9816: I can switch to ```private```, but not ```final``` because it is initialized in a method and not in the constructor. If I do this the compilation fails.

gehel: Right, I missed that. We should move initialization to the ctor... that validateSetup method is a legacy left over that should disappear at some point. That being said, this is good enough to be merged. I'll merge tonight unless you want to do this additional cleanup.

nlaurent9816: I don't think I'll commit this evening. You can merge.

Thanks for all the comments! This is my first pull request on a open source project and i'm grateful to you for the reviews!

gehel:
@nlaurent9816 Thanks a lot for the patch! I hope this won't be your last open source contribution!

Issue #636, opened 2018-01-12
jmxtrans docker image


asiekkowa:
Could you fix the errors with jmxtrans image? https://hub.docker.com/r/jmxtrans/jmxtrans/builds/

It would be great if images had the up to date code - I try to avoid pull the code and build this image locally by myself.

gehel:
This docker image is fairly experimental, and does not receive the amount of love that it deserves. I'm afraid I don't have the time to make sure it is in good shape. But if anyone wants to take care of it, the jmxtrans team always welcomes contributions!

PR #668, opened 2018-04-23
verify downloaded jar file and suppress errors when deleting $GNUPGHOME

Telling:
This should fix https://github.com/jmxtrans/jmxtrans/issues/636 and furthermore verifies the downloaded jar against the pgp signature.


gehel:
@zonArt, would you mind having a look at this PR?

PR #681, opened 2018-06-11
InfluxDB : Allow string values and more

nlaurent9816:
Continuation of the work from #660 with the segregateStringValues parameter (renamed allowStringValues) which allows the InfluxDBWriter to write string fields in InfluxDB.

Additional work was also done: 
- Fix 2 bugs with the typeNames parameter
- Update the tests
- Add description of the InfluxDBWriterFactory parameters

I'm also working on a new parameter for the InfluxDBWriter, called "attributesAsTag": 
In InfluxDB, you can store values as tag or as fields. Tags are indexed, fields not. (See [here](https://docs.influxdata.com/influxdb/v1.5/concepts/key_concepts/).) Actually, the InfluxDbWriter stores every attributes value as fields. If this format is adapted for metrics (counters, averages, etc...), you can have performance issues when reading if you try to filter with them. Tags are more adapted for filtering, that's why it is recommended to store IDs, names, metadata, etc... as tags. Actually, only JMX port (with reportJmxPortAsTag parameter), typeName, objDomain, className, attributeName (with resultTags parameter) and full custom tags are written. "attributesAsTag" will be a list of attributes to write as tags instead of fields.
I haven't yet pushed the commits of "attributesAsTag" because I prefer to wait your opinion on this.

### DIFF
--- a/jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
+++ b/jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
@@ -22,7 +22,6 @@
  */
 package com.googlecode.jmxtrans.model.output;
 
-import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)


gquintana:
LGTM
@hightea Can you have a look at this PR?

hightea:
LGTM too
maybe add an assertion on the String value in the test (with escaped quotes)

nlaurent9816:
Thank you for your answer, @gquintana and @hightea.
I would also have your advice on the “attributeAsTag” feature (see the first comment). I started to code this on [my fork](https://github.com/nlaurent9816/jmxtrans/tree/attributesAsTags). Are you interested by such a feature? If yes, do you prefer I merge it on this PR or I create a new one?
I also notice the unused variable “isNotNaN” (InfluxDbWriter, line 85). I think I will remove it, unless you hase a specific reason to keep it.

gquintana:
I'll merge this PR as soon as I'll have figured out what the problem with snyk.io is

PR #686, opened 2018-07-05
InfluxDBWriter : Add "attributesAsTag" parameter

nlaurent9816:
In InfluxDB, you can store values as tag or as fields. Tags are indexed but can only be Strings, fields are not indexed but suuports more data formats. (See [here](https://docs.influxdata.com/influxdb/v1.5/concepts/key_concepts/).) Actually, the InfluxDbWriter stores every attributes value (results from queries) as fields. If this format is adapted for metrics (counters, averages, etc...), you can have performance issues when reading if you try to filter with them. Tags are more adapted for filtering, that's why it is recommended to store IDs, names, metadata, etc... as tags. Actually, only JMX port (with reportJmxPortAsTag parameter), typeName, objDomain, className, attributeName (with resultTags parameter) and full custom tags are written. "attributesAsTag" is a list of attributes to write as tags instead of fields.

I also have a question: How can I also edit the Wiki pages in order to document features?


nlaurent9816:
2 months after, anybody wants to check this PR?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
```diff
@@ -181,6 +181,8 @@ public synchronized void start() throws LifecycleException {
 				this.startupWatchdir();
 
 				this.startupSystem();
+				
+				this.registerMBeans();
```
gquintana: Why? How is it related to InfluxDB?

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -68,13 +69,15 @@
...
 	@Nonnull private final String retentionPolicy;
 	@Nonnull private final ImmutableMap<String,String> tags;
-	@Nonnull ImmutableList<String> typeNames;
-
 	/**
 	 * The {@link ImmutableSet} of {@link ResultAttribute} attributes of
 	 * {@link Result} that will be written as {@link Point} tags
 	 */
-	private final ImmutableSet<ResultAttribute> resultAttributesToWriteAsTags;
+	@Nonnull private final ImmutableSet<ResultAttribute> resultAttributesToWriteAsTags;
```
gquintana: tags, resultAttributesToWriteAsTags, typeNamesAsTags, attributesAsTags...
I wonder whether it would be cleaner to have a `List<TagSupplier>`, TagSupplier being a poorly named functional interface to extract tags from server/query/result.

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -161,44 +166,65 @@ public InfluxDbWriter(
...
 		ImmutableList<String> typeNamesParam = null;
 		// if not typeNamesAsTag, we concat typeName in values.
 		if (!typeNamesAsTags) {
 			typeNamesParam = this.typeNames;
 		}
 		
+		//Creates a tag map, associated by the typeName.
+		//Results fields with the same typeName will receive the same tag values.
+		HashMap<String,HashMap<String, String>> attrTagByTypeName = newHashMap();
+		if (!attributesAsTags.isEmpty()) {
```
gquintana: Extracting a method may make it more readable.
Is it possible to convert `HashMap<HashMap>` into a `Map<Map>`?

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
```diff
@@ -227,5 +253,9 @@ public void doWrite(Server server, Query query, Iterable<Result> results) throws
 		return resultTagMap;
 
 	}
+	
+	private boolean isValidValue(Object value) {
+		return isValidNumber(value) || allowStringValues && value instanceof String;
```
gquintana: There are 2 changes in this PR: attributesAsTags and allowStringValues
Can you split this PR?

nlaurent9816: allowStringValues was already in a independant PR : #681 
It was already merged early July.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
nlaurent9816: Oh, it's a commit pushed in the wrong branch. It's not related to InfluxDB, but it fixes a bug : when you modify/delete a file in the config folder, jmxtrans reloads the config. But when it do that, it tries to register some MBeans alerady registered, and the process stops without restarting the server jobs.

How can I cancel this commit on github?

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
nlaurent9816: No problem, it will be done in the next commit.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
nlaurent9816: Ok, it's now cancelled.

jmxtrans-output/jmxtrans-output-influxdb/src/main/java/com/googlecode/jmxtrans/model/output/InfluxDbWriter.java
nlaurent9816: I'll see if I can do something like that.

PR #704, opened 2018-09-21
Fix race condition in UdpLoggingServer for ITests

gquintana:


gehel:
Good try! We should probably ditch stormpot and go back to commons-pool, easier to manage (not entirely related to this issue, but would probably make it easier to address).

PR #708, opened 2018-10-10
Package JMXTrans as a web application

gquintana:
The idea is to package JMXTrans as a War which can be deployed in old school application servers like WebLogic, Tomcat, JBoss... 
This allows to monitor theses application servers from the inside without opening RMI ports.

We are using successfully, 

The only problem is the dependency of JMXTrans on JCommander to load base configuration. I tried several refactoringss but they brought me to break all the things! This should be done in a separate PR.


gehel:
[embedded-jmxtrans](https://github.com/jmxtrans/embedded-jmxtrans) might be a better fit for this (at least easier), but having some convergence between the different jmxtrans flavours would be really nice as well!

gehel:
This definitely has my +1! Looks like a very good idea! We refactor the configuration as you pointed out already.

I'm wondering what should be the minimal requirements here. Should we remove some of that XML and go directly for newer @WebListener?

gquintana:
* I ignored there was an embedded-jmxtrans. How many jmxtrans flavours are there?
* I am using this on old WebLogic 10.x which is not Servlet 3.0 compliant :'( not @WebListener here

gehel:
> * I ignored there was an embedded-jmxtrans. How many jmxtrans flavours are there?

Far too many! jmxtrans (standalone), jmxtrans-embedded, jmxtrans-agent. See https://github.com/jmxtrans/

> * I am using this on old WebLogic 10.x which is not Servlet 3.0 compliant :'( no `@WebListener` here

You're probably not the only one. Let's be conservative!

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/cli/JmxTransConfiguration.java
```diff
@@ -144,10 +169,138 @@ public File apply(String input) {
...
+			if (value != null) {
+				doSetValue(value, configuration);
+			}
+		}
+
+		protected abstract void doSetValue(T value, JmxTransConfiguration configuration);
+	}
+
+	private static PropertySetter[] SETTERS = new PropertySetter[]{
+			new PropertySetter<Boolean>(CONTINUE_ON_ERROR_PROPERTY, Boolean.class) {
```
gehel: This would be soo much cleaner with lambda. Is anyone still using Java 7 (I'm loosing touch with enterprise realities)

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/cli/JmxTransConfiguration.java
```diff
@@ -144,10 +169,138 @@ public File apply(String input) {
...
+			setter.setValue(typedProperties, this);
+		}
+	}
+
+	public void loadProperties(File configFile) throws IOException {
+		Properties properties = new Properties();
+		try (InputStream in = this.getClass().getClassLoader().getResourceAsStream("jmxtrans.properties")) {
+			properties.load(in);
+		}
+		if (configFile != null) {
```
gehel: We could probably reduce duplication and increase readability by extracting a method to find the right file to load and one to load it.

gquintana: This does property merging, you can override just some properties.
The order of priority is now (from most important to less important):

1. Command line args
2. Provided properties file or if not provided and exists default jmxtrans.properties
3. jmxtrans.properties in classpath
4. Default values in JmxtransConfiguration

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/cli/JmxTransConfiguration.java
```diff
@@ -144,10 +169,138 @@ public File apply(String input) {
...
+
+	@VisibleForTesting
+	void loadProperties(Properties properties) {
+		TypedProperties typedProperties = new TypedProperties(properties);
+		for(PropertySetter setter: SETTERS) {
+			setter.setValue(typedProperties, this);
+		}
+	}
+
+	public void loadProperties(File configFile) throws IOException {
```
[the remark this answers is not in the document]
gquintana: I'll introduce a JmxtransConfigurationFactory with static methods.

I thought as you that JmxTransConfiguration could be made immutable, but JmxtransConfiguration can be modified by ManagedJmxTransformerProcess so we can't remove setters. I suspect JCommander uses them as well.

This brings a question: why are there 2 implementations of `CliArgumentParser` one based on JCommander, the other on Commons Cli?

gquintana:
@gehel These snyk.io checks are a bit painful, because we can't see what the problem is.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/cli/JmxTransConfiguration.java
gquintana: I saw a problem with sortpom plugin and Java 7 on my machine, I wonder whether it stills compile on Java 7

gehel:
> @gehel These snyk.io checks are a bit painful, because we can't see what the problem is.

I've removed snyk.io

gquintana:
@gehel Can I merge?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/cli/JmxTransConfiguration.java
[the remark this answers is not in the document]
gehel: That's probably a left over from the migration from Commons Cli to JCommander. I don't think Commons Cli is used anymore, we could probably remove it.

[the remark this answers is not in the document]
gehel: Note that the whole configuration parsing is still ugly and should really be cleaned up! If you think it is a mess, I agree!

gehel:
Yes, please merge!

(unrelated, but we should probably have a chat one day about releasing jmxtrans, you should be able to do it without my involvement when you need to. Ping me if interested - guillaume.lederrey@gmail.com or gehel on IRC/freenode).

PR #709, opened 2018-10-10
Terminate jmxtrans when it runs into an exception

afn:
Address one common cause of uncaught exceptions, and also register an uncaught exception handler to cause jmxtrans to exit with a non-zero status rather than getting stuck in an inconsistent state.


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
```diff
@@ -492,6 +505,22 @@ private void deleteAllJobs() throws Exception {
...
+		this.uninitializeExecutors(queryExecutorRepository);
+		this.uninitializeExecutors(resultExecutorRepository);
+	}
+
+	private void uninitializeExecutors(ExecutorRepository executorRepository) {
+		executorRepository.clear();
+	}
+
+	private void restartSystem() throws Exception {
+		this.deleteAllJobs();
```
gehel: I'm surprised by all those explicit `this.`. It looks somewhat coherent with the rest of the code in this class, but this is (I hope) not how all the code in this project looks. (Very minor comment, feel free to ignore it).

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
```diff
@@ -492,6 +505,22 @@ private void deleteAllJobs() throws Exception {
 		}
 	}
 
+	private void uninitializeExecutors() throws MalformedObjectNameException {
+		this.uninitializeExecutors(queryExecutorRepository);
+		this.uninitializeExecutors(resultExecutorRepository);
+	}
+
+	private void uninitializeExecutors(ExecutorRepository executorRepository) {
```
gehel: I'm not entirely sure that extracting `uninitializeExecutors(ExecutorRepository executorRepository)` added a lot of clarity, maybe inlining it would be better. (really minor, feel free to ignore)

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
afn: Ha, I usually don't `this` when it's not needed, but I saw a lot of it in this file so I followed suit :) I'm adding a commit that removes *all* unnecessary `this` from this file.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
afn: I was on the fence about this one but it seemed to make sense to have parity between the `initializeExecutors` and `uninitializeExecutors` methods. That way, if in the future there's some additional initialization happening, it'll be clear where to put the corresponding teardown code.

afn:
:+1: I've rebased this against `ignore-hidden-files` so this should merge without conflicts after #710 is merged.

afn:
Thanks for the quick review! Once #709, #710, and #712 have been merged, would you mind putting out a new release?

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
```diff
@@ -159,7 +172,7 @@ public static void main(String[] args) throws Exception {
 	 */
 	private void doMain() throws Exception {
 		// Start the process
-		this.start();
+		start();
```
gquintana: There are many unneeded changes, mostly code style, in this PR. Removing `this.` is not needed and adds unnecessary noise to this PR. Maybe move this to another PR.

jmxtrans-core/src/main/java/com/googlecode/jmxtrans/JmxTransformer.java
```diff
@@ -550,8 +579,7 @@ private void scheduleReload() {
 			@Override
 			public void run() {
 				try {
-					deleteAllJobs();
-					startupSystem();
+					restartSystem();
 				} catch(Exception e) {
 					throw new RuntimeException(e);
```
gquintana: Does it make sense, to move the try-catch in the restartSystem method and remove the throws Exception from signature

afn: This code is actually changed in a subsequent PR (#714); throwing a RuntimeException isn't the right thing to do here anyway.

afn:
Is this good to merge, or anything else that needs to be addressed?

ankitchiplunkar:
This is an important PR, when will it be merged?

jpzk:
Yea, we also have some issues with this problem in our infrastructure. Please make this part of the next release!

gquintana:
This PR is very close to https://github.com/jmxtrans/jmxtrans/pull/731
Does it fix the same problem?

jeroenr:
Please merge this! We're running jmxtrans as a sidecar to our kafka & mirrormaker pods, but they often end up in "zombie mode" after exceptions like mentioned in this thread. I had to work around it by creating a liveness probe which monitors outgoing network traffic to detect when it stopped writing to our backend (influx). Messy at best!

coykitten:
Is there anything that can be done to help get this over the finish line? We're running into this issue as well.

PR #712, opened 2018-10-10
Minor improvements to jmxtrans-systemd

afn:
* Fixes to jmxtrans-systemd logback configuration:    
    1. Fix error in how `logDir` property was defined
    2. Remove console appender. No need to log redundantly to both the system journal and the log file.
* Allow jmxtrans service to be enabled to start on boot: In order for `systemd enable jmxtrans` to cause it to start on boot, the `WantedBy` attribute needs to be specified.


afn:
Rebased against master.

gquintana:
LGTM @afn can you rebase please?

PR #713, opened 2018-10-11
Relax JDK dependency in Debian package

afn:
The system java installation may come from a different package than the ones listed, so make it optional (with `Recommended`) rather than a strict dependency.

### DIFF
--- a/jmxtrans-packaging/jmxtrans-systemd/src/deb/control/control
+++ b/jmxtrans-packaging/jmxtrans-systemd/src/deb/control/control
@@ -4,6 +4,6 @@ Maintainer: Guillaume Lederrey <guillaume.lederrey@gmail.com>
 Architecture: all
 Section: admin
 Priority: optional
-Depends: java7-runtime-headless | openjdk-7-jre-headless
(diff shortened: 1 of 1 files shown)


afn:
Sorry @gehel, got one more for you :) In our systems we had to install the package with `--ignore-depends=java7-runtime-headless`, which would have been an acceptable workaround except that all future `apt-get` invocations complain about the missing dependency.

gehel:
Never be sorry for improving a project!

In this case, I suppose that your JRE is not installed via a .deb package, and so apt does not know about it. I would probably argue that the fix should be in the way you install your JRE ([equivs](https://manpages.debian.org/testing/equivs/equivs-build.1.en.html) might be your friend here). jmxtrans will not run without a JRE 7 (or later), so declaring a Depends instead of a Recommends seems correct to me.

I'm happy to continue this discussion, I'm definitely not an expert in Debian packaging!

afn:
That's a good point. So I've changed the PR to leave it as `Depends` but be more lenient in what packages it allows: rather than strictly requiring java 7, it will also accept `default-jre-headless` or `java8-runtime-headless`.

afn:
Closing this PR, since as you pointed out, newer Java packages will `Provides` the java7-runtime-headless dependency

PR #714, opened 2018-10-11
Handle errors on reload with the default uncaught exception handler

afn:
This code (added in #711) was throwing a RuntimeException, which I mistakenly believed would invoke the default uncaught exception handler. Instead, ThreadPoolExecutor swallows the RuntimeException and terminates the executor.

Calling the default uncaught exception handler will currently just log the exception to stderr. Once #709 is merged, it will trigger our global uncaught exception handler, which logs the error and shuts down jmxtrans.


afn:
Rebased.

PR #735, opened 2018-12-19
Skip separator when key prefix is empty

sopel39:
It is often convenient not to have any key prefix but only attribute name.

For instance when storing metrics in CloudWatch Agent via statsd instead of having long metric name:
`coordinator_presto_memory_NonHeapMemoryUsage_used`
I would rather have:
`NonHeapMemoryUsage_used`
as I can identify metric source via so called "dimensions" (statsd attributes).


gquintana:
LGTM

PR #736, opened 2018-12-19
Add support for keys filtering to jmx result processor

sopel39:
Keys filtering is useful in cloud environments where one pays
price per metric. In some cases, (e.g: `NonHeapMemoryUsage`) some
keys have meaningless values, so it doesn't make sense to evaluate them
for every machine.

### DIFF
--- a/jmxtrans-core/src/test/java/com/googlecode/jmxtrans/jmx/JmxResultProcessorTest.java
+++ b/jmxtrans-core/src/test/java/com/googlecode/jmxtrans/jmx/JmxResultProcessorTest.java
@@ -28,6 +28,7 @@ import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.googlecode.jmxtrans.model.JmxResultProcessor;
+import com.googlecode.jmxtrans.model.Query;
 import com.googlecode.jmxtrans.model.Result;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


sopel39:
Supersedes: https://github.com/jmxtrans/jmxtrans/pull/316

sopel39:
@gquintana Could you review this too?

jmxtrans-core/src/test/java/com/googlecode/jmxtrans/jmx/JmxResultProcessorTest.java
```diff
@@ -237,13 +238,17 @@ public void canReadCompositeData() throws MalformedObjectNameException, Attribut
 		AttributeList attr = ManagementFactory.getPlatformMBeanServer().getAttributes(
 				memory.getObjectName(), new String[]{"HeapMemoryUsage"});
 		List<Result> results = new JmxResultProcessor(
-				dummyQueryWithResultAlias(),
+				Query.builder()
```
gquintana: Can you add a test rathen modifying an existing one? Can we have both tests with and without keys?

sopel39: Added separate test

sopel39:
@gquintana Applied comment. Could you take a look again?

sopel39:
ping @gquintana

gquintana:
Thanks for contributing

PR #737, opened 2018-12-27
[SensuWriterFactory] Added the possibility to configure metric name, type and handler

magnuslarsen:
Beside using a consistent tab-indentation and splitting some arguments to span multi-lines, the only thing i changed was added the ability to configure:
* Metric name (string: name, default value: "metric-jmxtrans")
* Metric type (string: type, default value: "metric")
* Metric handler (string: handler, default value: "graphite")

### DIFF
--- a/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/SensuWriterFactory.java
+++ b/jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/SensuWriterFactory.java
@@ -48,40 +48,56 @@ public class SensuWriterFactory implements OutputWriterFactory {
 	@Nonnull private final InetSocketAddress server;
 	@Nonnull private final ImmutableList<String> typeNames;
 	@Nullable private final String rootPrefix;
+	@Nonnull private final String handler;
+	@Nonnull private final String name;
+	@Nonnull private final String type;
 	@Nonnull private final FlushStrategy flushStrategy;
 	private final int poolSize;
 
 	public SensuWriterFactory(
-			@JsonProperty("typeNames") ImmutableList<String> typeNames,
-			@JsonProperty("booleanAsNumber") boolean booleanAsNumber,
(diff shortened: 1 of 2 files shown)


jmxtrans-output/jmxtrans-output-core/src/main/java/com/googlecode/jmxtrans/model/output/SensuWriterFactory.java
```diff
@@ -48,40 +48,56 @@
...
-			@JsonProperty("booleanAsNumber") boolean booleanAsNumber,
-			@JsonProperty("host") String host,
-			@JsonProperty("port") Integer port,
-			@JsonProperty("rootPrefix") String rootPrefix,
-			@JsonProperty("flushStrategy") String flushStrategy,
-			@JsonProperty("flushDelayInSeconds") Integer flushDelayInSeconds,
-			@JsonProperty("poolSize") Integer poolSize) {
-		this.rootPrefix = rootPrefix;
-		this.typeNames = firstNonNull(typeNames, ImmutableList.<String>of());
+		@JsonProperty("booleanAsNumber") boolean booleanAsNumber,
```
gquintana: Why did you reorder arguments?
Except that LGTM

magnuslarsen: Ordering is now back to normal, with my 3 additions added at the bottom, in the order they are called in (handler -> name -> type)

magnuslarsen:
After the reordering the build fails, seemingly because `UdpOutputWriterBuilderIT` test fails, with a timeout. I cannot see why reverting the ordering in `SensuWriterFactory`should make the `UdpOutputWriterBuilderIT` slower, as it imports the `TcpOutputWriterBuilder`, not Udp.
Should i increase the timeout value, or do you have any suggestions?

gquintana:
LGTM

gquintana:
Could you rebase this PR please?

PR #738, opened 2019-01-02
added GetAlias type definition

magnuslarsen:
Added the type GetAlias(), which is used by KeyUtils.java (most likely in more places):
https://github.com/jmxtrans/jmxtrans/blob/712b24f4f9e162b8456a15eea87517b5fff9cb66/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/naming/KeyUtils.java#L97-L106

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
@@ -376,6 +376,14 @@ public class Server implements JmxConnectionProvider {
 		return url.substring(url.lastIndexOf("//") + 2, url.lastIndexOf(':'));
 	}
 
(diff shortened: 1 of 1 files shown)


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -376,6 +376,14 @@ public String getHost() {
 		return url.substring(url.lastIndexOf("//") + 2, url.lastIndexOf(':'));
 	}
 
+	public String getAlias() {
```
gquintana: No need to check whether alias is null

magnuslarsen: Removed.

PR #767, opened 2019-12-22
Print simple error message when addition jar on Java 9+

gquintana:
SystemClassLoader is not of type URLClassLoader,
as a result additional jar feature doesn't work on recent Java versions.
Fix the error message in that situation.

I wonder whether we shouldn't remove the additional jar option as well.


gehel:
Some application servers require additional classes to be added to the classpath for JMX. Having the ability to add them via a CLI option instead of tuning the classpath manually is a nice shortcut for some users. That being said, it is ugly. Removing it in favor of good documentation, or a higher level wrapper might be a good idea.

PR #780, opened 2020-08-20
fix NPE when a result is Null

hfutatzhanghb:
When an attribute in JmxResultProcessor#attributes  is null, it throws NPE and leads to other attributes missing。So，we should judge whether the current attribute is null or not. If current attribute is null, just ignore it.


hfutatzhanghb:
In my company's practice, this commit is useful.

PR #797, opened 2021-03-30
Keep outputwriters in server copy

### LINKED ISSUE #796: In Server.build(server), the already builded outputwriters are not copied
The Server.build(server) method doesn't copy the outputWriters from original server.
This causes the loss of the outputwriters when they are defined at server level in a server without explicit runPeriodSeconds.

(caused by this additionhttps://github.com/jmxtrans/jmxtrans/blob/058df191036421ddaf2d8d4797ca4de2515639d0/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/scheduler/ServerScheduler.java#L87 )

A workaround is to define a runPeriodSeconds at server level

hightea:
Copy the builded outputwriters when building a server from an other one.

Fixes #796

Note that in SeverListBuilder.TemporaryServer.build we clean this copied list in order to use the singleton value of the outputwriters

### DIFF
--- a/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
+++ b/jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
@@ -31,9 +31,7 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
 import com.google.inject.name.Named;
-import com.googlecode.jmxtrans.ServerListBuilder;
 import com.googlecode.jmxtrans.connections.JMXConnection;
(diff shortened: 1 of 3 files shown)


jmxtrans-core/src/main/java/com/googlecode/jmxtrans/model/Server.java
```diff
@@ -508,6 +510,7 @@ private Builder(Server server) {
 			this.local = server.local;
 			this.ssl = server.ssl;
 			this.queries.addAll(server.queries);
+			Iterables.addAll(this.outputWriters, server.outputWriters);
```
gquintana, line 513: ` this.outputWriters.addAll(server.outputWriters)`  is enough, is it?

hightea: Done. Doing it, I was wondering if we could use simple Set type for those collections (`Server.outputWriters` And `Server.outputWriterFactories`). But it could be done later in a simplification effort.

PR #800, opened 2021-08-18
Remove the jpathwatch dependency and just use the nio classes.

jonathan-albrecht-ibm:
This PR removes the dependency on jpathwatch and replaces it with the standard nio implementation.

jmxtrans has a dependency on [jpathwatch](https://jpathwatch.wordpress.com/) which is used to watch directories for file changes. jpathwatch contains jni code and native libs for a few x86 platforms. It provides a pure java fallback for the rest but logs a lot of exceptions about the missing native libs which makes it look like there is a problem. jpathwatch also has not been updated since 2013.

Since Java 1.7, file watching is part of the standard Java runtime. Now that jmxtrans specifies Java 1.8, we should be able to switch to the standard Java classes and drop the dependency on jpathwatch.

### DIFF
--- a/jmxtrans-utils/pom.xml
+++ b/jmxtrans-utils/pom.xml
@@ -53,10 +53,6 @@
 			<artifactId>commons-lang</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>net.sf.jpathwatch</groupId>
-			<artifactId>jpathwatch</artifactId>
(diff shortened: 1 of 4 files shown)


jonathan-albrecht-ibm:
gentle nudge: Just wondering if there is interest in this PR? Happy to help in any way I can.

